PR #34, opened 2015-07-10
Negative tracking

### LINKED ISSUE #33: Fix negative tracking
```objc
if ( trackingInPoints > 0.0f ) {
```

This line in `BONTextConfiguration` precludes negative tracking values, which should be supported. This should be changed to `trackingInPoints != 0.0f` or, to appease the float comparison gods, a custom `floatsCloseEnough(trackingInPoints, 0.0f)`.

ZevEisenberg:
Fixes #33


ZevEisenberg:
floating point math strikes again (this fails only on 32-bit).

ZevEisenberg:
Well, that turned out to be less trivial than I was expecting. 😅

PR #41, opened 2015-07-17
#2 special characters

### LINKED ISSUE #2: Make it easy to use special characters
To use characters like non-breaking thin spaces and other difficult-to-type-and-also-to-see symbols, you usually have to consult a Unicode reference and then include an obtuse Unicode code point in your string. BonMot should make it easy to add special characters to strings by providing a convenient API to access these characters. It should be easy to add these characters in the context of concatenating attributed strings. For example, you might want to make a string like this:

```
[image][non-breaking thin space]Some Text[thin space][image][non-breaking thin space]More Text
```

This typically takes multiple join/concatenate passes, with obscure Unicode characters or wordy abstractions. BonMot should make the code to generate strings like this concise and readable.

ZevEisenberg:
Fixes #2 

cc @ateliercw thought you might be interested to see this. Places to look are `BONSpecial` for implementation and `SpecialCharactersCell` for usage. Also `BONSpecialGenerator.swift` for the ugly, ugly generation script.

- [x] need to update the readme


scripts/BONSpecialGenerator.swift
```diff
@@ -0,0 +1,158 @@
...
+
+// ******************************
+// *                            *
+// *  Real program starts here  *
+// *                            *
+// ******************************
+
+// Make sure specialCharacters is sorted
+let sortedSpecialCharacters = specialCharacters.sort({$0 < $1})
+if sortedSpecialCharacters != specialCharacters {
```
ateliercw: #? is this just to make sure no dupes are in the set? Couldn't you cast the initial array as a set, and sort that rather than worrying about if the input is sorted unless you're fielding a pull request?

ateliercw:
Have you run the swift script through swiftlint? It tends to be brutal, but make you think about how to make things more swift-y

I really like how the appending syntax looks for appending special characters.

At a glance, no print special chars in debug strings yet?

scripts/BONSpecialGenerator.swift
ZevEisenberg: It handles dupes, but it also makes sure they're also in the same order as in the generated files, which makes it easier to debug and edit the file.

ZevEisenberg:
@ateliercw `-debugDescription` of `BONText` does print special characters. It's one character per line, so it could use some cleanup, but I'll do that separately.

PR #65, opened 2015-08-24
32 carthage support

mattThousand:
Built `BonMot` as a framework w/ a public target. Thanks goes out to @ateliercw. 

Still need to test, but should work by adding the following to any `Cartfile` and running `carthage update`:

```
github "Raizlabs/BonMot" "e72ca55a110a9932ce4dc4b9bf4eb8245d8c4e95"
```


ZevEisenberg:
For this to work in the future, does the .xcodeproj need to be updated whenever a class is added or deleted? If so, are there ways that we can run the test suite using both CocoaPods and Carthage, to make sure they're both configured correctly? We could add a Rakefile if it would make this easier.

ateliercw:
the .xcodeproj does need to be updated if files are added or deleted, the simplest test would just be adding checking to make sure that the .xcodeproj builds, as it will fail to build if anything is referenced in a header but not included.

ZevEisenberg:
@ateliercw @mattThousand I re-built the project from a clean `pod lib create BonMot` using CocoaPods 0.39.0, and it included a new `_Pods.xcodeproj` file, which is there for Carthage compatibility. Please have a look at the project as it is set up in `master` and let me know if that satisfies the requirements for using Carthage.

ateliercw:
@ZevEisenberg as soon as your latest tagged release includes the .xcodeproj it looks like bonmot will be carthage compatible (it built fine with carthage when pointed at the latest commit sha)

ateliercw:
@ZevEisenberg - just tested carthage on 2.1.1 and it builds bonmot

ZevEisenberg:
:tada:

PR #76, opened 2015-10-21
fixes the crash due to range calcualting longer than string range

### LINKED ISSUE #75: BonText attributedStringLastConcatenant: crashes on zero length strings
With a BonChain that includes a zero length, but non-null, `string` property, BonMot crashes when attempting to generate an attributed string for the chain.

Submitting a patch to fix the issue

ateliercw:
Fixes #75


ZevEisenberg:
@ateliercw could you please include a failing test in your PR?

PR #88, opened 2015-11-11
add special character - horizontal ellipsis

mr-fixit:
add '...'
add generator script to project to ease finding it.


Example/BonMot/BONSpecialGenerator.swift
```diff
@@ -0,0 +1,219 @@
...
+    0x2003,
+    0x2007,
+    0x2009,
+    0x200A,
+    0x200B,
+    0x2011,
+    0x2012,
+    0x2013,
+    0x2014,
+    0x2026, // HORIZONTAL ELLIPSIS
```
ZevEisenberg: #req No need to comment, unless you want to go through and comment all of them. I prefer to just look up the appropriate code in the .m file if you need it. (Commenting each one could be nice, though - what do you think?)

Example/BonMot/BONSpecialGenerator.swift
```diff
@@ -0,0 +1,219 @@
+#!/usr/bin/env swift
```
ZevEisenberg: It looks like you've duplicated the generator script. If you're going to reference it in the project (which is not a bad idea), it should either a) be put in as a file reference, or b) moved on the file system. Also, this gave me the idea for https://github.com/Raizlabs/BonMot/issues/89

Example/BonMot/BONSpecialGenerator.swift
mr-fixit: Commenting helps a little here, but as you say, one could just look in .m file.  I have no strong feelings either way.

If I were to put on my "let's over-engineer-this" hat, I could suggest that the comment could be some form that would allow the script to sanity-check the name it derives against the name that the author intended.  But it's not a hat day.

ZevEisenberg: untitled
That's just one more thing to get out of sync. The .m file is law - if it's right, pretty much nothing else matters. And that array is kept sorted in code just so it's easy to find where to insert new characters later.

Example/BonMot/BONSpecialGenerator.swift
mr-fixit: I think I fixed that just now.  I deleted the version under Examples, and referenced the one in ./scripts.

ZevEisenberg: I still see a copy at `BonMot/Example/BonMot/BONSpecialGenerator.swift`.

Example/BonMot.xcodeproj/project.pbxproj
```diff
@@ -155,6 +155,7 @@
 		CDFDF5501B51880D00A1AD42 /* NSDictionary+BONEquality.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = "NSDictionary+BONEquality.m"; sourceTree = "<group>"; };
 		CDFDF5531B518F3900A1AD42 /* BONDictionaryEqualityTestCase.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = BONDictionaryEqualityTestCase.m; sourceTree = "<group>"; };
 		DC26879CFC50F3E9302E163B /* Pods-Tests.debug.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-Tests.debug.xcconfig"; path = "Pods/Target Support Files/Pods-Tests/Pods-Tests.debug.xcconfig"; sourceTree = "<group>"; };
+		E19B7EA11BF3C517002CD3D8 /* BONSpecialGenerator.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BONSpecialGenerator.swift; sourceTree = "<group>"; };
```
ZevEisenberg: #req this is still referencing the one from the Example project.
screen shot 2015-11-11 at 2 22 22 pm

ZevEisenberg:
Thanks @mr-fixit!

Issue #14, opened 2015-05-21
Underlining


ZevEisenberg:
Don't bother implementing this unless someone specifically requests it for a shipping product, and even then, try to talk them out of it. No one likes underlining except my high school English teachers.

lifely:
Hello, i'm sure you're gonna talk me out of it but could I specifically request this ? :x
 
It's not that i like this or not, it's just that specs are specs and sometime you don't choose what you have to do.

Also as a wrapper around attributedStrings shouldn't BonMot handle all it can do ? Maybe if you don't explicitly implement it, having a way to add attributes to `BONChain` object would be appreciated.

ZevEisenberg:
Definitely! Specs, as you so rightly point out, are specs :D

If you’d like to take a crack at a PR, I’d be happy to provide help if you need it. It should be pretty similar to the way font and color are implemented, if you need a stating point.

ZevEisenberg:
Oh, and I agree that eventually BonMot should support everything that attributed strings do. Definitely a medium-term goal, after basic usability improvements and a better Swift interface.

ZevEisenberg:
Going to do separate properties for underline style and underline color, in case you want to specify the style but stick with the default text color. The color part is easy, because it behaves just like the existing `color` and `backgroundColor` properties that BonMot already has.

`NSUnderlineStyle`, however, is weird:

```objc
// This defines currently supported values for NSUnderlineStyleAttributeName and NSStrikethroughStyleAttributeName. NSUnderlineStyle*, NSUnderlinePattern*, and NSUnderlineByWord are or'ed together to produce an underline style.
typedef NS_ENUM(NSInteger, NSUnderlineStyle) {
    NSUnderlineStyleNone                                    = 0x00,
    NSUnderlineStyleSingle                                  = 0x01,
    NSUnderlineStyleThick NS_ENUM_AVAILABLE(10_0, 7_0)      = 0x02,
    NSUnderlineStyleDouble NS_ENUM_AVAILABLE(10_0, 7_0)     = 0x09,

    NSUnderlinePatternSolid NS_ENUM_AVAILABLE(10_0, 7_0)      = 0x0000,
    NSUnderlinePatternDot NS_ENUM_AVAILABLE(10_0, 7_0)        = 0x0100,
    NSUnderlinePatternDash NS_ENUM_AVAILABLE(10_0, 7_0)       = 0x0200,
[...]

Issue #15, opened 2015-05-21
Support strikethrough text


ZevEisenberg:
BonMot should make it easy to strike through part or all of a string. If multiple strikethrough styles are available, those should be supported as well.

lifely:
:+1:  :shipit:

ZevEisenberg:
Closed via https://github.com/Raizlabs/BonMot/pull/102

PR #98, opened 2015-11-29
Implemented first proposition talked in issue #14, i think using defa…

lifely:
Hey BonMot Team,

So after proposing implementing underlining feature in BonMot on issue #14, and being suprised by the response on the issue : ).

I gave it a go and implemented it, as discussed in the issue, there was two implementation possible, i chose the first one and implemented it using the system-types because most of BonMot already does it that way.

Also I believe that using something simple and that people know is better.
If you are an accustomed objective-c dev you would guess that `NSUnderlineStyle` with a `BONChain.new.underlineStyle(NSUnderlineStyleSingle)` would work out the box and try it.

At least that's what i tried quickly w/ that cool framework.

Finally, in the issue @ZevEisenberg mentioned a `underlineByWord`, i think that feature could still be possible and I could add it, if you wanted to.

This is some cool work, really helps a lot in the day-to-day basics.
Can't wait to see more project like this flourish in our community.

### DIFF
--- a/Pod/Classes/BONChain.m
+++ b/Pod/Classes/BONChain.m
@@ -210,6 +210,28 @@
     return [indentSpacerBlock copy];
 }
 
+- (BONChainUnderlineStyle)underlineStyle
+{
+  BONChainUnderlineStyle underlineStyleBlock = ^(NSUnderlineStyle style) {
+    __typeof(self) newChain = self.copyWithoutNextText;
+    newChain.text.underlineStyle = style;
+    return newChain;
+  };
+
+  return [underlineStyleBlock copy];
+}
+
+- (BONChainUnderlineColor)underlineColor
+{
+  BONChainUnderlineColor underlineColorBlock = ^(UIColor *color) {
+    __typeof(self) newChain = self.copyWithoutNextText;
+    newChain.text.underlineColor = color;
+    return newChain;
+  };
+
+  return [underlineColorBlock copy];
+}
+
 - (void)appendLink:(id<BONChainable>)link
 {
     [self appendLink:link separator:nil];
--- a/Pod/Classes/BONText.h
+++ b/Pod/Classes/BONText.h
@@ -66,6 +66,9 @@ typedef NS_ENUM(NSUInteger, BONFigureSpacing) {
  */
 @property (nonatomic) CGFloat indentSpacer;
 
+@property (assign, nonatomic) NSUnderlineStyle underlineStyle;
+@property (strong, nonatomic) UIColor          *underlineColor;
+
 // Getting Values Out
 
 @property (copy, nonatomic, readonly) NSAttributedString *attributedString;
--- a/Pod/Classes/BONText.m
+++ b/Pod/Classes/BONText.m
@@ -285,6 +285,16 @@ static inline BOOL BONCGFloatsCloseEnough(CGFloat float1, CGFloat float2)
         attributes[NSParagraphStyleAttributeName] = paragraphStyle;
     }
 
+    // Text Underlining
+
+    if (self.underlineStyle != NSUnderlineStyleNone) {
+        attributes[NSUnderlineStyleAttributeName] = @(self.underlineStyle);
+    }
+
+    if (self.underlineColor != nil) {
+        attributes[NSUnderlineColorAttributeName] = self.underlineColor;
+    }
+
     return attributes;
 }
 
@@ -305,7 +315,12 @@ static inline BOOL BONCGFloatsCloseEnough(CGFloat float1, CGFloat float2)
     text.string = self.string;
     text.image = self.image;
     text.nextText = self.nextText;
+
+    // !!!: why are you not copying indentSpace too
     text.internalIndentSpacer = self.internalIndentSpacer;
+
+    text.underlineStyle = self.underlineStyle;
+    text.underlineColor = self.underlineColor;
 
     return text;
 }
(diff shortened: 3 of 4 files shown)


ZevEisenberg:
@lifely thanks for your PR! I agree with you and @KingOfBrian that the `NSUnderlineStyle` approach is better than individual home-grown attributes.

I'm going to provide some feedback on your PR using the following key:
#req - Required, i.e. this PR will not be merged until this is fixed. 
#nth - Nice to have, i.e. this would be nice to have now, but will not block merge. Fix eventually. 
#pp - Personal preference, i.e. "I would have done it this way, but this is acceptable"
#? - Question, i.e. "Why did you do it this way?"

Pod/Classes/BONChain.m
```diff
@@ -210,6 +210,28 @@ - (BONChainIndentSpacer)indentSpacer
     return [indentSpacerBlock copy];
 }
 
+- (BONChainUnderlineStyle)underlineStyle
+{
+  BONChainUnderlineStyle underlineStyleBlock = ^(NSUnderlineStyle style) {
```
ZevEisenberg: #req please use four spaces to indent. See the [Contributing](https://github.com/Raizlabs/BonMot#contributing) section of the ReadMe if you want to use `clang-format` to double check the formatting.

Pod/Classes/BONText.h
```diff
@@ -66,6 +66,9 @@ typedef NS_ENUM(NSUInteger, BONFigureSpacing) {
  */
 @property (nonatomic) CGFloat indentSpacer;
 
+@property (assign, nonatomic) NSUnderlineStyle underlineStyle;
```
ZevEisenberg: #req don't specify `assign`. This is technically a #pp, but one I feel very strongly about. Since the advent of ARC, `assign` is pretty much never a good choice for objects, and is the only option for primitives. Therefore, if I see `assign` in a project, I immediately assume that it should not be there.

Pod/Classes/BONText.m
```diff
@@ -285,6 +285,16 @@ - (NSDictionary *)attributes
         attributes[NSParagraphStyleAttributeName] = paragraphStyle;
     }
 
+    // Text Underlining
+
+    if (self.underlineStyle != NSUnderlineStyleNone) {
```
ZevEisenberg: #nth I know `NSUnderlineStyleNone == 0` in the current implementation, and is unlikely to change, but I would feel better if this were set explicitly in `-init`.

Pod/Classes/BONText.m
```diff
@@ -285,6 +285,16 @@ - (NSDictionary *)attributes
...
+        attributes[NSUnderlineStyleAttributeName] = @(self.underlineStyle);
+    }
+
+    if (self.underlineColor != nil) {
```
ZevEisenberg: #pp I typically don't use an explicit `nil` check in this project, preferring instead `if (self.underlineColor) {` for object `nil` checks. But I know that Swift requires an explicit check, which I like, so the only argument against doing it here is consistency with the rest of the code base.

Pod/Classes/BONText.m
```diff
@@ -305,8 +315,13 @@ - (id)copyWithZone:(NSZone *)zone
     text.string = self.string;
     text.image = self.image;
     text.nextText = self.nextText;
+
+    // !!!: why are you not copying indentSpace too
```
ZevEisenberg: #answer `indentSpacer` is backed by `internalIndentSpacer`:

```objc
- (CGFloat)indentSpacer
{
    return self.internalIndentSpacer.doubleValue;
}

- (void)setIndentSpacer:(CGFloat)indentSpacer
{
    self.internalIndentSpacer = @(indentSpacer);
}
```

This is so that we can distinguish between a value of `0.0` and a value of `nil`, which should have different behaviors.

ZevEisenberg:
strong #nth: It would be great to have some unit tests for this stuff. The unit tests in BonMot can be a bit opaque, so let me know if you want help or can think of a cleaner way to do it. They're mostly tests to make sure that the attributes that you expect are applied to the correct ranges, and especially since you can mix and match different underline styles, patterns, and colors, I think tests would be a good idea. I'm currently working to get test coverage data integrated into CI, so that this comment wouldn't be necessary.

ZevEisenberg:
#nth it would be great to knock out #15 (strikethroughs) as well in this PR, since they're so similar to underlining. But since you specifically requested underlining, it's fine if you don't want to add it yourself.

ZevEisenberg:
@lifely overall, this looks great. Will merge if you clean up the formatting and delete your `// !!!` comment (if I've answered to your satisfaction). Everything else is optional.

ZevEisenberg:
One more #req: please add `- Underlining` (and `Strikethrough`, if you implement it) to the `Supported Text Features` section of the ReadMe.

lifely:
Okay, @ZevEisenberg,

I've tried sticking to your formatting style but apparently i've missed some spaces, il fix that asap.

For the other thing you've mentioned, i agree mostly, i'll see if i can find some time to work on the Strikethrough sometime early this week and i'll update the Readme once all that done too.

ZevEisenberg:
@lifely bump. Any progress with this? I can take it over if you don't plan on finishing it.

ZevEisenberg:
@lifely closing in favor of #102, which builds on your initial work. I didn't want to rush you, but I had a free hack day and I wanted to get this merged. Thanks for your contribution! Look for a 2.2 release within the hour.

PR #100, opened 2015-12-15
Update README to show Carthage compatibility

nuudles:


README.md
```diff
@@ -18,6 +19,14 @@ it, simply add the following line to your Podfile:
 pod 'BonMot'
 ```
 
+## Installation with Carthage
+
+BonMot is also compatible with [Carthage](https://github.com/Carthage/Carthage). To install it, simply add the following line to your Cartfile:
+
+```ogml
```
ZevEisenberg: This should be `ogdl` (I keep making this same mistake).

ZevEisenberg:
@mattThousand @ateliercw I don't use Carthage. If one of you has a sec, can you confirm that this is what we need to say in order to claim Carthage compatibility?

ateliercw:
@ZevEisenberg it matches the copy used by other carthage compatible packages

mattThousand:
that's the gist of it, you might add a line about running `carthage update`

ZevEisenberg:
Going to keep it minimal. We don't say how to run `pod install` either. Merging. @nuudles thanks, and congratulations on being the first non-in-house developer to land a pull request in BonMot!

nuudles:
Thanks @ZevEisenberg ! I've got more coming too, stay tuned.

PR #104, opened 2015-12-28
Add more paragraph style attributes

nuudles:
This commit adds several more attributes from the NSParagraphStyle properties, including the firstLineHeadIndent, headIndent, tailIndent, maximumLineHeight, minimumLineHeight, paragraphSpacing, and paragraphSpacingBefore properties. Unit tests have also been added to test these new properties.

### DIFF
--- a/Pod/Classes/BONChain.h
+++ b/Pod/Classes/BONChain.h
@@ -17,8 +17,15 @@ typedef BONChain * (^BONChainFont)(UIFont *font);
 typedef BONChain * (^BONChainColor)(UIColor *color);
 typedef BONChain * (^BONChainAdobeTracking)(NSInteger adobeTracking);
 typedef BONChain * (^BONChainPointTracking)(CGFloat pointTracking);
+typedef BONChain * (^BONChainFirstLineHeadIndent)(CGFloat firstLineHeadIndent);
+typedef BONChain * (^BONChainHeadIndent)(CGFloat headIndent);
+typedef BONChain * (^BONChainTailIndent)(CGFloat tailIndent);
 typedef BONChain * (^BONChainLineHeight)(CGFloat lineHeightMultiple);
+typedef BONChain * (^BONChainMaximumLineHeight)(CGFloat maximumLineHeight);
+typedef BONChain * (^BONChainMinimumLineHeight)(CGFloat minimumLineHeight);
 typedef BONChain * (^BONChainLineSpacing)(CGFloat lineSpacing);
+typedef BONChain * (^BONChainParagraphSpacing)(CGFloat paragraphSpacing);
+typedef BONChain * (^BONChainParagraphSpacingBefore)(CGFloat paragraphSpacingBefore);
 typedef BONChain * (^BONChainBaselineOffset)(CGFloat baselineOffset);
 typedef BONChain * (^BONChainAlignment)(NSTextAlignment alignment);
 typedef BONChain * (^BONChainFigureCase)(BONFigureCase figureCase);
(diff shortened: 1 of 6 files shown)


ZevEisenberg:
@nuudles :heart: :heart: :heart: for the time you're putting into this! I'm going to be [very busy](http://cherylandzev.com) until just after the new year, so I won't be able to give this a good look until then - sorry for the wait. Two things jump out after a quick peek, though:

1. It would be nice to differentiate between `paragraphSpacingBefore` and `paragraphSpacingAfter`. Apple's API isn't as clear as it could be, but we can use this opportunity to make it better.
1. BonMot's `indentSpacer` property uses `headIndent` internally to do its thing. Have you looked at cases where this might conflict with your `headIndent` property? Try it out in `IndentationCell.m` in the example project and see if you run into any interesting edge cases worth testing or throwing exceptions for.

nuudles:
@ZevEisenberg Congrats dude! No worries on any delay, I know what it's like to plan one of those. I'll take a look at your initial feedback.

Cheers!

ZevEisenberg:
@nuudles this all looks great, aside from my earlier comments.

nuudles:
@ZevEisenberg Sorry for the delay in taking a look at this. Work got away from me. I've updated the name of `paragraphSpacing` to `paragraphSpacingAfter`. As far as the `headIndent` property, I spent some time with playing around with the `indentSpacer` property with `headIndent` and it seems that the `indentSpacer` always overwrites `headIndent`, which is probably as expected. Do you think we should add some documentation to that effect though so that the behavior is documented?

ZevEisenberg:
@nuudles A little inline documentation would be appreciated. And if you've got the time, a unit test to make sure that behavior stays consistent. Thanks for getting back to this - always appreciated!

Pod/Classes/BONChain.h
```diff
@@ -17,8 +17,15 @@ typedef BONChain * (^BONChainFont)(UIFont *font);
...
 typedef BONChain * (^BONChainAdobeTracking)(NSInteger adobeTracking);
 typedef BONChain * (^BONChainPointTracking)(CGFloat pointTracking);
+typedef BONChain * (^BONChainFirstLineHeadIndent)(CGFloat firstLineHeadIndent);
+typedef BONChain * (^BONChainHeadIndent)(CGFloat headIndent);
+typedef BONChain * (^BONChainTailIndent)(CGFloat tailIndent);
 typedef BONChain * (^BONChainLineHeight)(CGFloat lineHeightMultiple);
+typedef BONChain * (^BONChainMaximumLineHeight)(CGFloat maximumLineHeight);
+typedef BONChain * (^BONChainMinimumLineHeight)(CGFloat minimumLineHeight);
 typedef BONChain * (^BONChainLineSpacing)(CGFloat lineSpacing);
+typedef BONChain * (^BONChainParagraphSpacing)(CGFloat paragraphSpacing);
```
ZevEisenberg: Please rename the block typedef to match. (Left this comment in the wrong place before)

nuudles:
@ZevEisenberg I've tried to add initial tests and documentation as requested regarding the `headIndent` and `indentSpacer` behavior. Unfortunately, I'll have limited time in the near future coming back to this if there are additional changes needed.

As a side note, going back to writing Obj C code after dealing with Swift for so long is harder than anticipated.

ZevEisenberg:
@nuudles thanks. I'll take it from here if any additional changes need to be made. And I agree about writing Obj-C. I've only been doing Swift in a hobby project, but it's completely ruined me for remembering to put `@` on `NSString` literals.

ZevEisenberg:
@nuudles I made a couple of tweaks (1aa95e392748ef3baac6acd6cab91a9ccb60a97f...f1dfff2f8e027f0bd8645475ed127ccd3dfa5606 - nothing major) and merged. Thanks for your contribution!

nuudles:
@ZevEisenberg That's great! Thanks for implementing the tweaks for me!

PR #113, opened 2016-03-07
Fix Xcode 7.3 warnings about float/double precision

ZevEisenberg:
@nbonatsakis @ateliercw @arrouse I don't know if anyone still cares about the Objective-C style guide, but this defnitely goes against it in the interest of cleaning up these warnings without just disabling the warnings in the project. This is so that people can use this in their projects without having to inhibit warnings from BonMot.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,6 @@
 
 language: objective-c
 osx_image: xcode7.2
-# cache: cocoapods
-# podfile: Example/Podfile
-# before_install:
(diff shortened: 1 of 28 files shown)


ZevEisenberg:
I split this into multiple commits and force-pushed. https://github.com/Raizlabs/BonMot/commit/7f52e11770b81a58deffaaa34beb2abef320cbf2 is the least boring of the commits.

Issue #40, opened 2015-07-16
Unit testing helpers


ZevEisenberg:
One of the intents of BonMot is to make it easy to write unit tests for the attributed strings that a view model might return. These strings may contain hard-to-distinguish characters such as nonbreaking spaces, which can make it difficult to write unit tests.

BonMot should have a method to transform a string into a human-readable version, like this:

```
{image attachment}{nonbreaking space}My cool label text{hair space}{image attachment}{nonbreaking space}Another piece of label text
```

The purpose would be to write a unit testing macro that works like this:

```objc
BONAssertEquivalentStrings(someViewModel.someProperty, @"{image attachment}{nonbreaking space}My cool label text{hair space}{image attachment}{nonbreaking space}Another piece of label text");
```

You would pre-generate the testing string via a utility method in BonMot, and then copy and paste it into your unit test. The `BONAssertEquivalentStrings()` macro takes two strings, runs the BonMot conversion on the first one, and then `XCTAssertEqualObjects`es them to see if they contain the same characters.

Bonus points: generate a useful error message that highlights, via a `^` character on the next line, at which index the two strings first differ.

PR #114, opened 2016-03-09
Issue #40 Create Macro to Test Properties Against a Human Readable String

eliotw1:
This is a first crack at addressing issue #40 

Comments & suggestions very welcome.


Pod/Classes/NSAttributedString+BonMotUtilities.m
```diff
@@ -0,0 +1,88 @@
+//
+//  NSAttributedString+BonMotUtilities.m
+//  Pods
```
ZevEisenberg: #req please replace `Pods` in your headers with `BonMot`. Xcode is annoying, about this, but I updated all the files to match recently, so it would be nice to keep them consistent.

Pod/Classes/NSAttributedString+BonMotUtilities.m
```diff
@@ -0,0 +1,88 @@
...
+            NSTextAttachment *attachment = attributes[NSAttachmentAttributeName];
+            UIImage *attachedImage = attachment.image;
+            NSString *imageSubstitutionString = [NSString stringWithFormat:@"{image%.0fx%.0f}", attachedImage.size.height, attachedImage.size.width];
+            [composedHumanReadableString appendString:imageSubstitutionString];
+        }
+        else {
+            // Substitute BONSpecial Characters with @"{#name}"
+            unichar character = [substring characterAtIndex:0];
+            BONGeneric(NSDictionary, NSNumber *, NSString *)*specialNames = @{
+                @(BONCharacterLineFeed) : @"{lineFeed}",
```
ZevEisenberg: #? any way to auto-generate this list using `BONSpecialGenerator.swift`, so that it will be updated when new characters are added to `BONSpecial`? Happy to discuss options here, since the generator script has some weird edge cases.

Pod/Classes/NSAttributedString+BonMotUtilities.m
```diff
@@ -0,0 +1,88 @@
...
+                @(BONCharacterNonBreakingHyphen) : @"{nonBreakingHyphen}",
+                @(BONCharacterFigureDash) : @"{figureDash}",
+                @(BONCharacterEnDash) : @"{enDash}",
+                @(BONCharacterEmDash) : @"{emDash}",
+                @(BONCharacterHorizontalEllipsis) : @"{horizontalEllipsis}",
+                @(BONCharacterLineSeparator) : @"{lineSeparator}",
+                @(BONCharacterParagraphSeparator) : @"{paragraphSeparator}",
+                @(BONCharacterNarrowNoBreakSpace) : @"{narrowNoBreakSpace}",
+                @(BONCharacterWordJoiner) : @"{wordJoiner}",
+                @(BONCharacterMinusSign) : @"{minusSign}"
```
ZevEisenberg: #pp trailing comma here, so that if another item is added later, it doesn't dirty this line in Git history.

Pod/Classes/NSAttributedString+BonMotUtilities.m
```diff
@@ -0,0 +1,88 @@
...
+            }
+            else if ([substring rangeOfCharacterFromSet:s_newLineCharacterSet].location != NSNotFound) {
+                // If it's a newline character, append a @"{newline}".
+                [composedHumanReadableString appendString:@"{newline}"];
+            }
+        }
+    }];
+
+    if (composedHumanReadableString.length == 0) {
+        [composedHumanReadableString appendString:@""];
```
ZevEisenberg: #? why append an empty string? Isn't it a no-op? If it's intentional, it should have a comment, because it looks like a mistake.

eliotw1: Good call, totally not needed. I will remove it and include a test for empty strings.

eliotw1:
Updated with a new script-created dictionary to translate the special characters to their human readable version.

Pod/Classes/BONSpecial.m
```diff
@@ -33,5 +33,29 @@ + (NSString *)wordJoiner { return @"\u2060"; }
 + (NSString *)minusSign { return @"\u2212"; }
 + (NSString *)objectReplacementCharacter { return @"\uFFFC"; }
 
++ (NSDictionary *)humanReadableStringDictionary {
```
ZevEisenberg: #req `{` on next line

scripts/BONSpecialGenerator.swift
```diff
@@ -51,6 +51,13 @@ let charactersRequiringFormatStrings: Set<unichar> = [
     0x0020,
 ]
 
+// These special characters are to be excluded from the human readable string dictionary
```
ZevEisenberg: #req period after this sentence

scripts/BONSpecialGenerator.swift
```diff
@@ -51,6 +51,13 @@ let charactersRequiringFormatStrings: Set<unichar> = [
...
+// They are handled outside the generated dictionary
```
ZevEisenberg: #? what exactly are you saying here? Please clarify.

scripts/BONSpecialGenerator.swift
```diff
@@ -51,6 +51,13 @@ let charactersRequiringFormatStrings: Set<unichar> = [
...
+let charactersToExcludeFromHumanReadableStringDictionary: Set<unichar> = [
+    0x0020,
+    unichar(NSAttachmentCharacter),
```
ZevEisenberg: #nth Technically, you could have an `NSAttachmentCharacter` without an image. I'm not saying it's likely, but there should at least be defined behavior and a unit test.

scripts/BONSpecialGenerator.swift
```diff
@@ -147,6 +154,8 @@ if sortedSpecialCharacters != specialCharacters {
 var headerEnumString = "typedef NS_ENUM(unichar, BONCharacter) {\n"
 var headerCodeString = ""
 var implementationCodeString = ""
+var humanReadableDictionaryHeaderDeclaration = "+ (NSDictionary *)humanReadableStringDictionary;"
```
ZevEisenberg: #pp declare these closer to where they're used
#req `let`, not `var`
#req for both header and implementation: this should use `BONGeneric` to expose a strongly typed dictionary.

Pod/Classes/NSAttributedString+BonMotUtilities.m
```diff
@@ -0,0 +1,61 @@
...
+            s_whiteSpaceAndNewLinesSet = [NSCharacterSet whitespaceAndNewlineCharacterSet];
+        }
+
+        BONGeneric(NSDictionary, NSNumber *, NSString *)*specialCharacterSubstitutionNameDictionary = [BONSpecial humanReadableStringDictionary];
+        unichar character = [substring characterAtIndex:0];
+        NSString *specialCharacterSubstitutionString = specialCharacterSubstitutionNameDictionary[@(character)];
+
+        // Substitute BONSpecial Characters with @"{#name}"
+        if (specialCharacterSubstitutionString) {
+            [composedHumanReadableString appendFormat:@"%@", specialCharacterSubstitutionString];
```
ZevEisenberg: #nth `appendString:`

Pod/Classes/NSAttributedString+BonMotUtilities.m
```diff
@@ -0,0 +1,61 @@
...
+            s_newLineCharacterSet = [NSCharacterSet newlineCharacterSet];
+        }
+        NSCharacterSet *s_whiteSpaceAndNewLinesSet = nil;
+        if (!s_whiteSpaceAndNewLinesSet) {
+            s_whiteSpaceAndNewLinesSet = [NSCharacterSet whitespaceAndNewlineCharacterSet];
+        }
+
+        BONGeneric(NSDictionary, NSNumber *, NSString *)*specialCharacterSubstitutionNameDictionary = [BONSpecial humanReadableStringDictionary];
+        unichar character = [substring characterAtIndex:0];
+        NSString *specialCharacterSubstitutionString = specialCharacterSubstitutionNameDictionary[@(character)];
```
ZevEisenberg: #pp this can be made less wordy with:

```objc
NSString *specialCharacterSubstitutionString = [BONSpecial humanReadableStringDictionary][@(character)];
```

Not required, and there are valid reasons to do it your way, but I think that would be my approach.

Pod/Classes/NSAttributedString+BonMotUtilities.m
```diff
@@ -0,0 +1,61 @@
...
+        if (specialCharacterSubstitutionString) {
+            [composedHumanReadableString appendFormat:@"%@", specialCharacterSubstitutionString];
+        }
+
+        // Substitute Images with @"{image#heightx#width}"
+        else if ([substring isEqualToString:BONSpecial.objectReplacementCharacter]) {
+            BONStringDict *attributes = [self attributesAtIndex:substringRange.location effectiveRange:NULL];
+            NSTextAttachment *attachment = attributes[NSAttachmentAttributeName];
+            UIImage *attachedImage = attachment.image;
+            NSString *imageSubstitutionString = [NSString stringWithFormat:@"{image%.0fx%.0f}", attachedImage.size.height, attachedImage.size.width];
```
ZevEisenberg: #? is it possible for an image to have a fractional size? If so, what are the possible values? And also if so, can/should we change this to reflect an integer number of pixels? I forget whether the image's `size` is in pixels or points.

Pod/Classes/BONSpecial.m
```diff
@@ -1,6 +1,6 @@
 //
 //  BONSpecial.m
-//  BonMot
+//  Pods
```
ZevEisenberg: #req please fix this in the generator script so it doesn't get reverted the next time someone runs it.

Pod/Classes/NSAttributedString+BonMotUtilities.h
```diff
@@ -0,0 +1,15 @@
...
+//
+//  Created by Eliot Williams on 3/9/16.
+//
+//
+
+#import <Foundation/Foundation.h>
+
+@interface NSAttributedString (BonMotUtilities)
+
+@property (nonatomic, readonly, nonnull) NSString *humanReadableString;
```
ZevEisenberg: #req category methods must be prefixed. Prefixing is annoying, and we could discuss other ways to accomplish this (a C function, for example, might make sense here), but if it's a category method (implicit with a property), it should be prefixed. In this case, I think the `bon_` prefix would actually make sense, because `humanReadableString` could mean different things in different contexts. I wonder whether a more specific name like "debuggingString` or similar might be more applicable?

Example/Tests/BONBaseTestCase.h
```diff
@@ -52,6 +53,11 @@ OBJC_EXTERN NSValue *BONValueFromRange(NSUInteger location, NSUInteger length);
         [mutableControlAttributes removeObjectForKey:testRangeValue]; }]; \
     XCTAssertEqual(mutableControlAttributes.count, 0, @"Some attributes not found in string: %@", mutableControlAttributes);
 
+#define BONAssertEquivalentStrings(attributedString, controlHumanReadableString)            \
+    NSAttributedString *castAttributedString = (NSAttributedString *)attributedString;      \
+    NSString *humanReadableAttributedString = castAttributedString.bon_humanReadableString; \
```
ZevEisenberg: #req wrong variable name

Example/Tests/BONHumanReadableStringTestCase.m
```diff
@@ -0,0 +1,85 @@
...
+    BONAssertEquivalentStrings(testAttributedStringProperty, expectedHumanReadableString);
+}
+
+- (void)testAllTheCharactersInTheKitchenSink
+{
+    BONChain *everything = BONChain.new;
+    [everything appendLink:BONChain.new.string(@"neon")];
+    [everything appendLink:BONChain.new.string(@"سلام")];
+    [everything appendLink:BONChain.new.string(@"🚲")];
+    [everything appendLink:BONChain.new.string(@"\U000A1337") separator:BONSpecial.figureDash];
```
ZevEisenberg: #nth/#? this character is good for the other test where it's used, but it actually reveals a problem with this string: it appears to be non-printing in this case, and its official name is `\N{<unassigned-A1337>}`. Is there any way to categorically detect that a character like this, and make it display in the human-readable string? If not, we should at least file a bug and use a different character for this test that makes the intent more clear. Or, leave this character in, so that the test fails if we ever fix it.

Example/Tests/BONIndentSpacerTestCase.m
```diff
@@ -7,6 +7,7 @@
 //
 
 #import "BONBaseTestCase.h"
+#import <BonMot/BonMot.h>
```
ZevEisenberg: #req you shouldn't need this

Pod/Classes/NSAttributedString+BonMotUtilities.m
```diff
@@ -0,0 +1,73 @@
...
+        CFStringTransform((CFMutableStringRef)mutableUnicodeName, NULL, kCFStringTransformToUnicodeName, FALSE);
+
+        BONStringDict *attributes = [self attributesAtIndex:substringRange.location effectiveRange:NULL];
+        NSTextAttachment *attachment = attributes[NSAttachmentAttributeName];
+        UIImage *attachedImage;
+        attachedImage = attachment.image;
+
+        // Substitute attached images with @"{image#heightx#width}"
+        if (attachedImage) {
+            NSString *imageSubstitutionString = [NSString stringWithFormat:@"{image%.0fx%.0f}", attachedImage.size.height, attachedImage.size.width];
```
ZevEisenberg: #req I believe the image could be a fractional number of points. That gets tricky, because a dimension could best be represented as `36` or `36.5` or `36.6666667` depending on the device and image size/scale factor. Thinking it should probably use a fixed number of decimal places. Also thinking it should have a `- (NSString *)bon_humanReadableStringIncludingImageSize:(BOOL)includeImageSize`, in case you're writing a unit test that needs to work with different image sizes.

Issue #89, opened 2015-11-11
Run BONSpecialGenerator.swift as part of every example app build


ZevEisenberg:
As a convenience, we could automatically run `BONSpecialGenerator.swift` when you build the example app. That would eliminate needing to go find and build it manually.

ZevEisenberg:
This would also catch bugs introduced into the generator script.

ZevEisenberg:
This is now fixed - thanks, @eliotw1!

PR #116, opened 2016-03-14
Issue #89: Run BONSpecialGenerator.swift before every build

eliotw1:
This will cause the script to be run more frequently, thus implementing any changes or errors.


ZevEisenberg:
Closing in favor of #126

PR #117, opened 2016-03-16
Change UI subspec to UIKit

ZevEisenberg:


Pod/Classes/BONText.h
```diff
@@ -88,6 +88,11 @@ typedef NS_ENUM(NSUInteger, BONFigureSpacing) {
 @property (copy, nonatomic, readonly, BONNonnull) NSAttributedString *attributedString;
 @property (copy, nonatomic, readonly, BONNonnull) BONStringDict *attributes;
 
+/**
+ *  @c YES if and only if the resulting @c attributedString will be the empty attributed string.
+ */
+@property (nonatomic, readonly) BOOL empty;
```
Imperiopolis: `empty` seems like a weird name for this property as I personally wouldn't consider a `BONText` empty if I'd assigned any of the properties. This feels like it should be named something more like `hasString`, though as discussed that is somewhat confusing and might be interpreted as `string != nil`.

ZevEisenberg: I'm going to call it `generatesEmptyString`. Considered making it private and calling it `hasEmptinessInside` but decided against it.

PR #118, opened 2016-03-16
Compilation on Non case folded filesystems will now work.

ejohansson:
Compilation on Non case folded filesystems will now work.


ZevEisenberg:
@ejohansson nice catch - thanks! I have no way of testing this easily, but I'm going to assume you have. Unfortunately, because some stuff from the next release has already been merged to master, this will have to wait to be pushed to trunk until 3.0, which will also contain minor breaking changes. Working to get it out soon (see other open PRs).

Issue #111, opened 2016-03-07
No support for lineBreakMode


tettoffensive:
It doesn't appear that BonMot has support for lineBreakMode so that you can set truncation to be ... among other options. Unless I am missing something. I can't find that this option is exposed.

ZevEisenberg:
@tettoffensive good catch! This is an accidental omission. If you're up for submitting a pull request, the changes from https://github.com/Raizlabs/BonMot/pull/101 and https://github.com/Raizlabs/BonMot/pull/104/files are essentially the same as what would need to change to support `lineBreakMode`.

tettoffensive:
Thanks. I'm up for it, but don't know if I'll get to it today.

ZevEisenberg:
No rush on my end. You're the first person who's requested it. Let me know if you're not going to be able to get to it; otherwise, I look forward to your PR :smile:

ZevEisenberg:
Hey @tettoffensive, is this still something you're interested in? I'm prepping for a 3.0 release that includes some breaking changes. It would be cool to get this in, if you're interested in working on it soon. If not, It doesn't introduce new breaking changes, so it could always go in later as a point release.

tettoffensive:
@ZevEisenberg I'll try to get this in!

tettoffensive:
@ZevEisenberg I made a PR here: https://github.com/Raizlabs/BonMot/pull/128

On another note, do you have a trick in xcode for your indentation with the BonChains? Where each `.` indents on a newline?

ZevEisenberg:
@tettoffensive thanks for the PR - I'll check it out!

As for indentation, Xcode gives me this for Swift:

```swift
attributedString = BONChain()
    .string("Hello, testing world")
    .font(UIFont.preferredFontForTextStyle(UIFontTextStyleBody))
    .textColor(UIColor.redColor())
    .attributedString
```

And this for Objective-C:

```objc
attributedString = BONChain.new
.string(@"Hello, testing world")
.font([UIFont preferredFontForTextStyle:UIFontTextStyleBody])
.textColor([UIColor redColor])
.attributedString;
```

So, no, my README indentations are lies, damned lies, I tell you! Sorry :disappointed:

ZevEisenberg:
@tettoffensive oops, forgot a wrinkle: for my Objective-C code, I use `clang-format` as outlined at the bottom of this project's README. It's what gives me the nice indentation like this:

```objc
NSAttributedString *attributedString =
[...]

PR #128, opened 2016-03-25
add support for line break mode

tettoffensive:
For https://github.com/Raizlabs/BonMot/issues/111

### DIFF
--- a/Example/Tests/BONTextAlignmentTestCase.m
+++ b/Example/Tests/BONTextAlignmentTestCase.m
@@ -75,10 +75,10 @@
                           .headIndent(2.34)
                           .tailIndent(3.45)
                           .lineHeightMultiple(3.14)
-                          .lineBreakMode(NSLineBreakByTruncatingTail)
                           .maximumLineHeight(5.67)
                           .minimumLineHeight(4.56)
                           .lineSpacing(2.72)
+                          .lineBreakMode(NSLineBreakByTruncatingTail)
                           .paragraphSpacingAfter(6.78)
                           .paragraphSpacingBefore(7.89);
     NSAttributedString *string = chain.attributedString;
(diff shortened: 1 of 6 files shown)


Pod/Classes/BONText.m
```diff
@@ -311,6 +311,13 @@ - (BONStringDict *)attributes
         populateParagraphStyleIfNecessary();
         paragraphStyle.lineSpacing = self.lineSpacing;
     }
+    
+    // Line Break Mode
+    
+    if (self.lineBreakMode != NSLineBreakByWordWrapping) {
```
ZevEisenberg: I would prefer not to rely on the implementation detail that this property is initialized to `0`, which happens to equal `NSLineBreakByWordWrapping`. Please add that to `-init` of `BONText.m`.

ZevEisenberg:
@tettoffensive thanks for your PR! If you'd be willing to address my one comment, then I'd happily test and merge this.

tettoffensive:
@ZevEisenberg happy to address. I've added to the init

Example/Tests/BONTextAlignmentTestCase.m
```diff
@@ -75,6 +75,7 @@ - (void)testParagraphStyle
                           .headIndent(2.34)
                           .tailIndent(3.45)
                           .lineHeightMultiple(3.14)
+                          .lineBreakMode(NSLineBreakByTruncatingTail)
```
ZevEisenberg: Would you mind moving this down below `.lineSpacing(2.72)`? Line break mode is after line spacing everywhere else, and that consistency makes it very easy to validate that it's in all the places it should be.

ZevEisenberg:
@tettoffensive just the one additional comment, and then I'm good to go. I double checked and tested it, and it looks great.

tettoffensive:
@ZevEisenberg moved the line

PR #131, opened 2016-04-02
Make BONText private

### LINKED ISSUE #127: Make BONText private
In daily usage, `BONChain` is the only way I've ever seen anyone actually use BonMot. `BONText` is a useful abstraction to keep the code clean, but it probably doesn't need to be public. Consider deprecating it, and removing it from the public headers in a future major release. If you disagree, and actually use `BONText` directly for any reasons, please post a comment - I'd love to hear from you!

ZevEisenberg:
Also used the opportunity to do a bunch of other smaller cleanup tasks.
Fixes #127


Pod/UIKit/Classes/UILabel+BonMotUtilities.m
```diff
@@ -7,33 +7,35 @@
...
-- (void)setBonTextable:(id<BONTextable>)textable
+- (void)setBonChain:(BONChain *)chain
 {
-    objc_setAssociatedObject(self, @selector(bonTextable), textable, OBJC_ASSOCIATION_COPY_NONATOMIC);
+    objc_setAssociatedObject(self, @selector(bonChain), chain, OBJC_ASSOCIATION_COPY_NONATOMIC);
 
-    // If the textable is empty, use the label’s existing text
-    if (textable.text.generatesEmptyString) {
+    // If the chain is empty, use the label’s existing text
+    if (chain.generatesEmptyString) {
```
KingOfBrian: This may be confusing behavior. If I intend to clear a label by setting an empty bonChain, won't this leave the last value in the label?

README.md
```diff
@@ -121,31 +103,31 @@ NSAttributedString *blueBirdString = blueBirds.attributedString;
...
-BONText *blueFish = BONChain.new.string(@"blue fish").textColor([UIColor blueColor]).text;
-BONText *separator = BONChain.new.string(@", ").text;
+BONChain *oneFish = BONChain.new.string(@"one fish");
+BONChain *twoFish = BONChain.new.string(@"two fish");
+BONChain *redFish = BONChain.new.string(@"red fish").textColor([UIColor redColor]);
+BONChain *blueFish = BONChain.new.string(@"blue fish").textColor([UIColor blueColor]);
+BONChain *separator = BONChain.new.string(@", ");
 
-NSAttributedString *string = [BONText joinTexts:@[ oneFish, twoFish, redFish, blueFish ] withSeparator:separator];
+NSAttributedString *string = [BONChain joinChains:@[ oneFish, twoFish, redFish, blueFish ] withSeparator:separator];
```
KingOfBrian: Might be nice to have a swift extension on CollectionType[BONChain] that implemented `chainsJoinedBySeparator()`.

Pod/UIKit/Classes/UILabel+BonMotUtilities.m
ZevEisenberg: That's a good point, though technically not part of this PR. @imperiopolis what do you think?

KingOfBrian:
Few small comments. Large +1 for the general simplification. One thought from the swift perspective would be to add initializers:

```obj-c
- (instancetype)initWithString:(NSString *)string
- (instancetype)initWithImage:(UIImage *)image
```

This might make more natural swifting, ala
```swift
let newStyle = BONChain(string:string)
let oldStyle = BONChain().string(string)
```
Looking at it, I'm not sure it's a big difference, but it's a bit more natural.

Also, on the Obj-C side, you could add:
```obj-c
+ (BONChainString)string;
+ (BONChainImage)image;
```
Which would let the API drop the `.new`. Again, not sure it's a big difference, but.

Pod/UIKit/Classes/UILabel+BonMotUtilities.m
Imperiopolis: Perhaps check if the old chain had a string that's equal to the text? If it did, clear the text. Otherwise, leave the text alone as it was likely assigned via one of the text properties directly.

README.md
ZevEisenberg: Might be, might be. With the new appending syntax, which you could also use with `reduce()`, I'm not finding that I really need the `joinChains` method as much as I thought I would, so it may become deprecated in a future release.

ZevEisenberg:
@KingOfBrian nice thought on the initializers, although the inconsistency of between sometimes using `(string: string)` and sometimes using `.string(string)` may not be worth the convenience. What if you want to add another link in the chain?

I much prefer your second suggestion, which helps a great deal in Swift, but would be super ungainly in Obj-C. I wonder if I can expose those methods only in Swift? Worst case, I think I can make an extension. Haven't worked with interop like that, so I've filed an issue to look into it: https://github.com/Raizlabs/BonMot/issues/132

Pod/UIKit/Classes/UILabel+BonMotUtilities.m
ZevEisenberg: It sounds a little too much like magic. I'm going to go with the most basic behavior with the fewest surprises: If you assign a chain, it always clobbers the text. I think this comports with how it would typically be used: assign the chain at setup, and then assign the string multiple times. If you're swapping out the chain, you can just mix in the string while you go. Otherwise, it's too complex to get what I see as the commonly desired behavior. This is going to be a breaking change compared to what we had originally discussed, @imperiopolis, but since this stuff hasn't shipped yet, I'm OK with it. I think @KingOfBrian makes a good point that you can't clear the label with an empty chain.

ZevEisenberg:
Per https://github.com/Raizlabs/BonMot/issues/127#issuecomment-209704081, I'm going to scrap this in favor of a less drastic move, where `BONText` is deprioritized in the ReadMe and the API, but still usable.

PR #137, opened 2016-04-14
De-emphasize BONText in ReadMe. Rename a method.

ZevEisenberg:


ZevEisenberg:
@ZevEisenberg from the future, would you mind reviewing this when you're not sleep-deprived?

PR #140, opened 2016-04-15
Code Coverage Reporting

ZevEisenberg:


ZevEisenberg:
This was just a test to see if I could get it working. Turns out: not easily enough to make it into this release. It's not a breaking change, so we can add it later.

PR #144, opened 2016-05-12
Add ability to add child BONChainables to style text between tags.

### LINKED ISSUE #69: Support styling localized strings

Imperiopolis:
This adds the ability to parse and style substrings between arbitrary tags with assigned `BONChainable` objects. This potentially resolves #69.

Excerpt from the readme:

## Tag Styles

BonMot can style text between arbirtrary tags using a `<tag></tag>` format and `\` as an escape character.

```objc
BONChain *boldChain = BONChain.new.fontNameAndSize(@"Baskerville-Bold", 15.0f);
BONChain *italicChain = BONChain.new.fontNameAndSize(@"Baskerville-Italic", 15.0f);

BONChain *chain = BONChain.new.fontNameAndSize(@"Baskerville", 17.0f)
    .tagStyles(@[BONTagMake(@"bold", boldChain), BONTagMake(@"italic", italicChain)])
    .string(@"<bold>This text is wrapped in a \\<bold> tag.</bold>\n<italic>This text is wrapped in an \\<italic> tag.</italic>");

NSAttributedString *string = chain.attributedString;
```

Outputs:

BonMot can also style text between any arbitrary start and end strings using any escape string.

```objc
BONChain *boldChain = BONChain.new.fontNameAndSize(@"Baskerville-Bold", 15.0f);
BONChain *italicChain = BONChain.new.fontNameAndSize(@"Baskerville-Italic", 15.0f);

BONChain *chain = BONChain.new.fontNameAndSize(@"Baskerville", 17.0f)
.tagStyles(@[BONTagComplexMake(@"~start~", @"!end", @"escape", boldChain)])
.string(@"~start~This text is wrapped in a escape~start~ tag.!end");

NSAttributedString *string = chain.attributedString;
```

**Note:** Tag styles do not support nested or interleaved tags. The first tag matched will be applied, any additional tags between the start end end will be ignored.


ZevEisenberg:
@Imperiopolis cool, thanks for posting this! I'm looking forward to digging in to it.

A question to start things off: why do you have the option to supply custom start, end and escape characters? Is it something you're using in your app? I ask because the syntax for `.tagStyles` could be simplified if you don't have to wrap tag/chain pairs in a constructor:

```objc
.tagStyles(@{@"bold": boldChain, @"italic": italicChain})
```

Of course, this looks better in Swift:

```swift
.tagStyles(["bold: boldChain, "italic": italicChain])
```

ZevEisenberg:
Also, taking a straight dictionary of `[String: BONTextable]` also makes it easier to compose things. Your app could have a base dictionary with bold, italics, etc., and then if you need a special style at a particular call site, you could do:

```swift
.tagStyles(Appearance.defaultStyles + ["specialCase": specialChain])
```

(Can you tell I'm ready to be done with Obj-C?)

Imperiopolis:
I initially wrote this exactly as you described, but as I was tidying things up I realized that the system was flexible enough to allow any arbitrary tag and it seemed worth exposing.

Some use cases I see for this:
* Parse tags that don't use `<>` – stuff like markdown style `*text*`, etc
* Allow for more complex tag handling – thinking stuff like `<span class="a-thing">text</span>`

For my particular uses, the former is not super useful but I intend to use the latter. I am utilizing this as a lightweight sort of HTML parser, such that my copy can essentially be shared between mobile and web with web using CSS styles and mobile using BonMot.

There are other libraries with more fully flushed out ways to handle complex HTML like this (utilizing CSS), but they all seem to incur a pretty significant performance hit to build out the full DOMTree. Since the styling I need to support is quite simple, I built out this lightweight version in BonMot to accommodate.

All that said, I think it could still be reasonably easy to accommodate your dictionary method by adding a function that takes the dictionary and converts it to the appropriate simple `BONTag` objects. I'm not sure I feel a solid need for doing that though. With this method, composition is still relatively straight forward.

```swift
.tagStyles(Appearance.defaultStyles + [BonTagMake("specialCase", specialChain)])
```

Will need to think on it a bit.

Imperiopolis:
I added in a `BONTagsFromDictionary` function which allows doing something like
```swift
.tagStyles(BONTagsFromDictionary(Appearance.defaultStyles + ["specialCase": specialChain]))
```

Also considering adding a `.tagStylesDictionary` property to allow for something like the following that automatically does the conversion and updates `.tagStyles`
```swift
.tagStylesDictionary(Appearance.defaultStyles + ["specialCase": specialChain])
```

ZevEisenberg:
> Also considering adding a `.tagStylesDictionary` property

I like this. Maybe make the default one, which takes a `[String : BONChain]` or `[String : BONTextable]`, be `.tagStyles`, and the advanced/complex one be `.advancedTagStyles` or `.customTagStyles` or `.complexTagStyles`? I'd prefer the basic API to look simple, rather than defined by the _absence_ of the complex API.

Imperiopolis:
That makes sense. I'll make that tweak :)

Example/BonMot/Cells/TagStylesCell.m
```diff
@@ -0,0 +1,40 @@
...
+
+- (void)awakeFromNib
+{
+    [super awakeFromNib];
+
+    BONChain *boldChain = BONChain.new.fontNameAndSize(@"Baskerville-Bold", 15);
+    BONChain *italicChain = BONChain.new.fontNameAndSize(@"Baskerville-Italic", 15);
+
+    BONChain *baseChain = BONChain.new.fontNameAndSize(@"Baskerville", 17)
+                              .tagStyles(@[ BONTagMake(@"bold", boldChain), BONTagMake(@"italic", italicChain) ])
```
ZevEisenberg: #req build warning: `Incompatible pointer types passing 'NSArray *' to parameter of type 'NSDictionary<NSString *,id<BONTextable>> * _Nullable'`

Pod/Classes/BONChain.m
```diff
@@ -353,6 +354,28 @@ - (BONChainStrikethroughColor)strikethroughColor
     return [strikethroughColorBlock copy];
 }
 
+- (BONTagStyles)tagStyles
+{
+    BONTagStyles tagStylesBlock = ^(NSDictionary *tagStyles) {
```
ZevEisenberg: #nth use generic declaration inside methods:
```objcBONTagComplexStyles tagStylesBlock = ^(BONGeneric(NSArray, BONTag *) *tagStyles) {
```
(also on next method)

Example/BonMot/Cells/TagStylesCell.m
```diff
@@ -0,0 +1,40 @@
...
+                              .string(@"<bold>This text is wrapped in a \\<bold> tag.</bold>\n<italic>This text is wrapped in an \\<italic> tag.</italic>");
```
ZevEisenberg: #req one of the main points of this feature is to be able to style substrings, and from viewing the example app, it's not obvious that this is the case, because each style is on its own line. It should be something like:

`"<bold>This text</bold> contains a \\<bold> tag.\n<italic>This text</italic> contains an \\<italic> tag."`

ZevEisenberg:
@Imperiopolis it would be nice if `<bold>This text</>` worked, or possibly even _any_ `</*>`. If you disagree, I'd be interested to hear why, since I'm new to parsers and grammars, and I may be missing things. I was thinking that `<bold>some bold text</> and much later, some <italic>italic</> text` would be a convenient way to write stuff.

Example/BonMot.xcodeproj/project.pbxproj
```diff
@@ -66,7 +64,12 @@
 		CD6DF05A1BF6B98500676E2D /* NSDictionary+BONEquality.m in Sources */ = {isa = PBXBuildFile; fileRef = CD6DF04F1BF6B53100676E2D /* NSDictionary+BONEquality.m */; };
 		CD78701E1CA5F8EC00B2714F /* EBGaramond12-Regular.otf in Resources */ = {isa = PBXBuildFile; fileRef = CD6DF03E1BF6AFAA00676E2D /* EBGaramond12-Regular.otf */; };
 		CDE658581C24ADD8009C7D09 /* BONUnderlineAndStrikethroughTestCase.m in Sources */ = {isa = PBXBuildFile; fileRef = CDE658571C24ADD8009C7D09 /* BONUnderlineAndStrikethroughTestCase.m */; };
+		EC2F199E1CEA324900D9F1FE /* BONTagDictionaryStylesTestCase.m in Sources */ = {isa = PBXBuildFile; fileRef = EC2F199C1CEA324300D9F1FE /* BONTagDictionaryStylesTestCase.m */; };
+		EC433DB31C88B7D9001B3ABE /* BONTagStylesTestCase.m in Sources */ = {isa = PBXBuildFile; fileRef = EC433DB11C88B65B001B3ABE /* BONTagStylesTestCase.m */; };
```
ZevEisenberg: #req rename the test case files to match what they're testing (they weren't updated when you changed the names of the tag style properties).

Example/Tests/BONTagComplexStylesTestCase.m
```diff
@@ -0,0 +1,406 @@
+//
+//  BONTagComplexStylingTestCase.m
+//  BonMot
+//
+//  Created by Nora Trapp on 3/3/16.
+//
```
ZevEisenberg: #nth a couple of these files are missing the copyright notice. I care more because of consistency than legality.

Example/BonMot/Cells/TagStylesCell.m
ZevEisenberg: ^ this results in a crash if you scroll down in the simulator

Pod/Classes/BONTag.h
```diff
@@ -0,0 +1,69 @@
...
+@protocol BONTextable;
+
+@interface BONTag : NSObject <NSCopying>
+
+@property (copy, nonatomic, readonly, BONNonnull) NSString *startTag;
+@property (copy, nonatomic, readonly, BONNonnull) NSString *endTag;
+@property (copy, nonatomic, readonly, BONNonnull) NSString *escapeString;
+@property (copy, nonatomic, readonly, BONNonnull) id<BONTextable> textable;
+
+- (BONNonnull instancetype)initWithStartTag:(BONNonnull NSString *)startTag endTag:(BONNonnull NSString *)endTag escapeString:(BONNonnull NSString *)escapeString textable:(BONNonnull id<BONTextable>)textable;
```
ZevEisenberg: #? any way to provide Swift versions of these with default parameters? I forget if that's possible, but it would certainly be nice if it is. You could also hide `BONTag*Make` from Swift in that case (again, if possible).

Pod/Classes/BONChain.h
```diff
@@ -13,7 +13,7 @@
 
 BON_ASSUME_NONNULL_BEGIN
 
-@class BONChain;
+@class BONChain, BONTag;
```
ZevEisenberg: #pp one per line

Pod/Classes/BONChain.h
```diff
@@ -94,6 +97,19 @@ typedef BONChain *BONCNonnull (^BONChainStrikethroughColor)(UIColor *BONCNullabl
 @property (copy, nonatomic, readonly) BONChainStrikethroughStyle strikethroughStyle;
 @property (copy, nonatomic, readonly) BONChainStrikethroughColor strikethroughColor;
 
+/**
+ *  Assign @p BONTextables to use in styling substrings surrounded in given tags.
```
ZevEisenberg: #nth `surrounded in` → `surrounded by`

Pod/Classes/BONChain.h
```diff
@@ -94,6 +97,19 @@ typedef BONChain *BONCNonnull (^BONChainStrikethroughColor)(UIColor *BONCNullabl
...
+ *  For example, ["b": boldTextable] would apply the @p boldChain
```
ZevEisenberg: #req `boldTextable` → `boldChain` (consistency)

Pod/Classes/BONTag.h
```diff
@@ -0,0 +1,69 @@
...
+ *
+ *  @return A @p BONTag instance representing the tag.
+ */
+CG_INLINE BONTag *BONCNonnull BONTagComplexMake(NSString *BONCNonnull startTag, NSString *BONCNonnull endTag, NSString *BONCNonnull escapeString, id<BONTextable> BONCNonnull textable)
+{
+    return [[BONTag alloc] initWithStartTag:startTag endTag:endTag escapeString:escapeString textable:textable];
+}
+
+/**
+ *  Create a tag using the default start and end formats and using \ as an escape character. <tag> </tag>
```
ZevEisenberg: #? why is `<tag> </tag>` at the end of this line?
#nth `@p \`

Pod/Classes/BONTag.h
```diff
@@ -0,0 +1,69 @@
...
+ *  Create a tag with a custom start/end/escape combination.
+ *
+ *  @param startTag     The start tag.
+ *  @param endTag       The end tag.
+ *  @param escapeString The escape string.
+ *  @param textable     The style to apply.
+ *
+ *  @return A @p BONTag instance representing the tag.
+ */
+CG_INLINE BONTag *BONCNonnull BONTagComplexMake(NSString *BONCNonnull startTag, NSString *BONCNonnull endTag, NSString *BONCNonnull escapeString, id<BONTextable> BONCNonnull textable)
```
ZevEisenberg: #? why `CG_INLINE` as opposed to, say, `NS_INLINE`?

Pod/Classes/BONTag.h
```diff
@@ -0,0 +1,69 @@
...
+ */
+CG_INLINE BONTag *BONCNonnull BONTagMake(NSString *BONCNonnull tag, id<BONTextable> BONCNonnull textable)
+{
+    return [[BONTag alloc] initWithTag:tag textable:textable];
+}
+
+/**
+ *  Create an array of tag objects from a dictionary of ["tag": textable]
+ *
+ *  @param dictionary A dictionary of textables mapped to tags.
```
ZevEisenberg: #req this is backwards: `A dictionary of tags mapped to textables.`

Pod/Classes/BONTag.m
```diff
@@ -0,0 +1,235 @@
...
+//  Pods
+//
+//  Created by Nora Trapp on 5/11/16.
+//
+//
+
+#import "BONTag_Private.h"
+#import "BONMot.h"
+
+static const NSString *kBONTagDefaultStartPrefix = @"<";
```
ZevEisenberg: #req the way these constants are defined, they can be redefined at run time without a compiler error. To get a compiler error if you try to reassign it, it should be:

```objc
static NSString * const kBONTagDefaultStartPrefix = @"<";
```

(`clang-format` may disagree about the spacing, which is fine.)

Imperiopolis:
> @Imperiopolis it would be nice if <bold>This text</> worked, or possibly even any </*>. If you disagree, I'd be interested to hear why, since I'm new to parsers and grammars, and I may be missing things. I was thinking that <bold>some bold text</> and much later, some <italic>italic</> text would be a convenient way to write stuff.

I see a few advantages to having an explicit close tag for all start tags:
* Allows for pseudo HTML parsing (things like `<strong></strong>`) which allows text to be shared across platforms
* Prepares for future support of nested tags
* Compared to a `</*>` version, drastically simplifies matching logic to not need to do any wildcard searching

That said, there's nothing *stopping* this method from being used as `<bold></>` and if you have a strong argument for it, I'm okay with changing the defaults to work that way. My gut feeling is these default should be HTML-esque, though.

ZevEisenberg:
Shared parsing across platforms is a good case. I don't think nested tags is, though: you can easily do nested tags with `</>`, which just closes the innermost tag. I rescind my idea to support `</*>`, because it sounds messy and potentially misleading.

How bad/slow would it make the code to support `</>`? I'm not super attached to the idea, however. Could be an added feature later, so feel free to ignore it.

Imperiopolis:
Yes, I suppose you can do nested tags with `</>`, I just really prefer the explicit nature of specific closed tags.

For example, it's pretty unclear when you drop a tag what's happening here:
`this is a <bold>bold thing with some <italic>bold italic</> stuff</> and regular stuff</>`

Which one of these was this meant to be?
this is a **bold thing with some _bold italic_ stuff** and regular stuff</>
this is a **bold thing with some _bold italic_** _stuff_ and regular stuff</>

What was the extraneous trailing `</>` for? Did I forget a start tag or..?

Where as with explicit tags, you can quickly realize that "oh, forgot the opening h2"
`this is a <bold>bold thing with some <italic>bold italic</italic> stuff</bold> and regular stuff</h2>`

And that this pretty clearly produces:
this is a **bold thing with some _bold italic_ stuff** and regular stuff</h2>


> How bad/slow would it make the code to support </>?

There is no performance impact switching to `</>` vs a more explicit `</bold>` – that impact was really only for the `</*>` route. This should just work if you use `</>` as the endTag on `BONTag`

Pod/Classes/BONTag.m
Imperiopolis: Ah, TIL – good to know! 👍

Pod/Classes/BONTag.h
Imperiopolis: Good question without any answer really. Can change to `NS_INLINE` if you prefer, what they do is identical.

Pod/Classes/BONTag.h
Imperiopolis: the `<tag> </tag>` at the end was meant to clarify what the "default start and end formats" are – any suggestions on how to make this clearer?

Example/BonMot/Cells/TagStylesCell.m
Imperiopolis: Ooops, forgot to update this one when I switched the default to take a dictionary.

ZevEisenberg:
@Imperiopolis OK, you've convinced me with the "makes editing harder" argument. Let's leave `</>` out for now. We can always add it as an enhancement later without breaking anything.

Pod/Classes/BONTag.h
ZevEisenberg: On a new line:

```
Example:
 @code
// code text
<tagName>Normal String</tagName>
<tagName>String with \<escaped characters\></tagName>
@endcode
```

^ or whatever would demonstrate escaping properly - which characters do you need to escape? Are those cases tested?

Syntax via http://stackoverflow.com/questions/19168423/what-are-the-new-documentation-commands-available-in-xcode-5

Pod/Classes/BONTag_Private.h
```diff
@@ -0,0 +1,49 @@
...
+/**
+ *  Finds all escaped tags within a given string.
+ *
+ *  @param string                The string to search.
+ *  @param tags                  The tags to search for.
+ *  @param stripEscapeCharacters If YES, the escape characters will be removed from the string and the resulting ranges.
+ *
+ *  @return An array of ranges representing the escaped tags.
+ */
++ (BONNonnull BONGeneric(NSArray, NSValue *) *)escapedRangesInString:(NSString *BONCNonnull *BONCNonnull)string withTags:(BONNonnull BONGeneric(NSArray, BONTag *) *)tags stripEscapeCharacters:(BOOL)stripEscapeCharacters;
```
ZevEisenberg: #req document the fact that `*string` is mutated to contain the version of the string with unescaped tags stripped.

Pod/Classes/BONTag.m
```diff
@@ -0,0 +1,235 @@
...
+    NSString *description = [NSString stringWithFormat:@"<%@: %p, startTag: %@, endTag: %@, textable: <%@: %p>>", NSStringFromClass(self.class), self, self.startTag, self.endTag, NSStringFromClass(self.textable.class), self.textable];
+
+    return description;
+}
+
+#pragma mark - Tag Matching
+
++ (NSArray *)escapedRangesInString:(NSString **)string withTags:(NSArray *)tags stripEscapeCharacters:(BOOL)stripEscapeCharacters
+{
+    NSMutableArray *escapedRanges = [NSMutableArray array];
```
ZevEisenberg: #nth `BONGeneric`

Pod/Classes/BONTag.h
```diff
@@ -0,0 +1,69 @@
...
+ *
+ *  @return An array of @p BONTag objects representing the tags in the dictionary.
+ */
+CG_INLINE BONGeneric(NSArray, BONTag *) * BONCNonnull BONTagsFromDictionary(BONGeneric(NSDictionary, NSString *, id<BONTextable>) * BONCNonnull dictionary)
+{
+    NSMutableArray *tags = [NSMutableArray array];
```
ZevEisenberg: #nth `BONGeneric`

Pod/Classes/BONTag.m
```diff
@@ -0,0 +1,235 @@
...
+- (NSString *)description
+{
+    NSString *description = [NSString stringWithFormat:@"<%@: %p, startTag: %@, endTag: %@, textable: <%@: %p>>", NSStringFromClass(self.class), self, self.startTag, self.endTag, NSStringFromClass(self.textable.class), self.textable];
+
+    return description;
+}
+
+#pragma mark - Tag Matching
+
++ (NSArray *)escapedRangesInString:(NSString **)string withTags:(NSArray *)tags stripEscapeCharacters:(BOOL)stripEscapeCharacters
```
ZevEisenberg: #req please search the project for `NSArray *` and replace with `BONGeneric`

Pod/Classes/BONText.m
```diff
@@ -89,11 +92,26 @@ - (NSAttributedString *)attributedStringLastConcatenant:(BOOL)lastConcatenant
...
+            rangesPerTag = [BONTag rangesInString:&string betweenTags:self.tagStyles stripTags:YES];
+        }
+
+        mutableAttributedString = [[NSMutableAttributedString alloc] initWithString:string
                                                                          attributes:self.attributes];
-        if (lastConcatenant && self.string.length > 0) {
-            NSRange lastCharacterRange = NSMakeRange(self.string.length - 1, 1);
+
+        for (BONTag *tag in rangesPerTag) {
+            NSDictionary *attributes = tag.textable.text.attributes;
```
ZevEisenberg: #req `BONStringDict`

Pod/Classes/BONTag.m
```diff
@@ -0,0 +1,235 @@
...
+{
+    NSMutableArray *escapedRanges = [NSMutableArray array];
+
+    NSString *theString = *string;
```
ZevEisenberg: #req even though this is declared as non-null, in Obj-C, it could still come in as null. Prefer asserting explicitly rather than randomly crashing.

Pod/Classes/BONTag.m
```diff
@@ -0,0 +1,235 @@
...
+
+    NSRange searchRange = NSMakeRange(0, theString.length);
+
+    // Iterate over the string finding each escape in order until there are no more escape strings
```
ZevEisenberg: #pp commas for clarity: `Iterate over the string, finding each escape in order, until there are no more escape strings`

ZevEisenberg:
@Imperiopolis is it intentional that you can have different escape characters for each tag in a scenario where you're using complex tags? What's the use case?

Pod/Classes/BONTag.m
```diff
@@ -0,0 +1,235 @@
...
+        if ([escapedRanges containsObject:[NSValue valueWithRange:stringRange]]) {
+            searchRange = NSMakeRange(NSMaxRange(stringRange), [stringToSearch length] - NSMaxRange(stringRange));
+            continue;
+        }
+
+        return stringRange;
+    }
+}
+
++ (NSArray *)rangesInString:(NSString **)string betweenTags:(NSArray *)tags stripTags:(BOOL)stripTags
```
ZevEisenberg: #? why would you ever not strip tags? You never pass `NO` here, in the tests or the example app.

Pod/Classes/BONTag.m
```diff
@@ -0,0 +1,235 @@
...
+        }
+
+        if (!nextTag) {
+            break;
+        }
+
+        NSRange range = NSMakeRange(nextEscapeRange.location, nextEscapedTagRange.length + nextEscapeRange.length);
+
+        if (stripEscapeCharacters) {
+            theString = [theString stringByReplacingOccurrencesOfString:nextTag.escapeString withString:@"" options:0 range:nextEscapeRange];
```
ZevEisenberg: #nth I believe this will be faster if you create a mutable string, mutate it a bunch inside the loop, instead of copying the string each time.

ZevEisenberg:
@Imperiopolis I may have missed something, but I don't think you test escaping the escape character anywhere. `\\` would be the simplest example of this, but I didn't see `\\\\`, which you would need to express that, in the source code.

Pod/Classes/BONTag.m
```diff
@@ -0,0 +1,235 @@
...
+            range = NSMakeRange(nextEscapeRange.location, nextEscapedTagRange.length);
+        }
+
+        [escapedRanges addObject:[NSValue valueWithRange:range]];
+
+        searchRange = NSMakeRange(NSMaxRange(range), [theString length] - NSMaxRange(range));
```
ZevEisenberg: #pp `theString.length` for idempotent getters

Pod/Classes/BONTag.m
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    *string = theString;
+
+    return escapedRanges;
+}
+
++ (NSRange)findNextString:(NSString *)string inString:(NSString *)stringToSearch ignoringRanges:(NSArray *)escapedRanges range:(NSRange)range
```
ZevEisenberg: #nth rename to `firstOccurrenceOfString:inString:ignoringRanges:inRange:`

ZevEisenberg: #nth these class methods are ripe for isolated testing. You can expose them in `BONTag_Testing.h` or in a category in the test header, if you want. Up to you, but it could help to clarify how they work.

Pod/Classes/BONTag.m
```diff
@@ -0,0 +1,235 @@
...
+
+- (BOOL)isEqualToTag:(BONTag *)tag
+{
+    BOOL startTagEqual = [tag.startTag isEqualToString:self.startTag];
+    BOOL endTagEqual = [tag.endTag isEqualToString:self.endTag];
+    BOOL chainEqual = [tag.textable isEqual:self.textable];
+    BOOL escapeEqual = [tag.escapeString isEqualToString:self.escapeString];
+    BOOL rangesEqual = [tag.ranges isEqualToArray:self.ranges];
+
+    return startTagEqual && endTagEqual && chainEqual && escapeEqual && rangesEqual;
```
ZevEisenberg: #nth string these together and order them smartly so you can take advantage of conditional short-circuiting, instead of calculating them all up front.

Pod/Classes/BONTag_Private.h
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ *  Search through a string to find the next matching string that is not in an escaped range.
+ *
+ *  @param string         The string to look for.
+ *  @param stringToSearch The string to search within.
+ *  @param escapedRanges  The ranges to ignore matches within.
+ *  @param range          The range of @p string to search within.
+ *
+ *  @return The range of the matched string. If the string is not found, location will be NSNotFound.
```
ZevEisenberg: #nth `@p NSNotFound`

Pod/Classes/BONTag_Private.h
```diff
@@ -0,0 +1,49 @@
...
+ */
++ (BONNonnull BONGeneric(NSArray, NSValue *) *)escapedRangesInString:(NSString *BONCNonnull *BONCNonnull)string withTags:(BONNonnull BONGeneric(NSArray, BONTag *) *)tags stripEscapeCharacters:(BOOL)stripEscapeCharacters;
+
+/**
+ *  Search through a string to find the next matching string that is not in an escaped range.
+ *
+ *  @param string         The string to look for.
+ *  @param stringToSearch The string to search within.
+ *  @param escapedRanges  The ranges to ignore matches within.
+ *  @param range          The range of @p string to search within.
```
ZevEisenberg: #pp `The range within @p string to search`

Pod/Classes/BONTag_Private.h
```diff
@@ -0,0 +1,49 @@
...
+ *  @return The range of the matched string. If the string is not found, location will be NSNotFound.
+ */
++ (NSRange)findNextString:(BONNonnull NSString *)string inString:(BONNonnull NSString *)stringToSearch ignoringRanges:(BONNonnull BONGeneric(NSArray, NSValue *) *)escapedRanges range:(NSRange)range;
+
+/**
+ *  Find all tagged strings within a given string.
+ *
+ *  @param string    The string to search.
+ *  @param tags      The tags to search for.
+ *  @param stripTags If YES, the start and end tags will be stripped from the string and the resulting ranges.
```
ZevEisenberg: #? what about escape sequences?

Pod/Classes/BONText.h
```diff
@@ -84,6 +84,11 @@ typedef NS_ENUM(NSUInteger, BONFigureSpacing) {
 @property (nonatomic) NSUnderlineStyle strikethroughStyle;
 @property (strong, nonatomic, BONNullable) UIColor *strikethroughColor;
 
+/**
+ *  Assign an array of @p BONTags to use in styling substrings.
```
ZevEisenberg: #nth describe in terms of what it _is_, now how you _use_ it.

ZevEisenberg: #pp `@p BONTag objects` so you end up with "`BONTag` objects`, and not "`BONTags`", in the formatted documentation.

Pod/Classes/BONText.m
```diff
@@ -89,11 +92,26 @@ - (NSAttributedString *)attributedStringLastConcatenant:(BOOL)lastConcatenant
             [mutableAttributedString appendAttributedString:[[NSAttributedString alloc] initWithString:@"\t" attributes:self.attributes]];
         }
     }
-    else if (self.string) {
-        mutableAttributedString = [[NSMutableAttributedString alloc] initWithString:self.string
+    else if (string) {
+        // If there is tag styling applied, strip the tags from the string and identify the ranges to apply the tag based chains to.
```
ZevEisenberg: #nth `tag-based chains`

README.md
```diff
@@ -267,6 +268,45 @@ Prints this:
 {image24x36}{noBreakSpace}Monday{enDash}Friday
 ```
 
+## Tag Styles
+
+BonMot can style text between arbirtrary tags using a `<tag></tag>` format and `\` as an escape character.
```
ZevEisenberg: #nth it would be nice to call out the (or at least, my) intended use of this feature: to parse localized strings.

ZevEisenberg:
@Imperiopolis OK, I’m done with my initial pass at the code review. Tiny tweaks may crop up, but in general it’s looking fantastic. Thanks so much for all the work you've put into this!

Pod/Classes/BONTag.h
Imperiopolis: You only need to escape tags that you've defined, so I'll do this as `\</tagName>`

Pod/Classes/BONTag.m
Imperiopolis: Good question, there really is no reason I can think of. Will remove this code path.

Imperiopolis:
> @Imperiopolis I may have missed something, but I don't think you test escaping the escape character anywhere. \\ would be the simplest example of this, but I didn't see \\\\, which you would need to express that, in the source code.

Escaping the escape character isn't something this currently supports – what's the goal of this? To handle if for some reason you needed a `\` right before a tag, but didn't want to escape?

Pod/Classes/BONTag_Private.h
Imperiopolis: Not sure what you're asking here?

Pod/Classes/BONText.m
Imperiopolis: Can you clarify?

README.md
Imperiopolis: Something like this?
> This allows applying styles to substrings within localized text where the position may change.

ZevEisenberg:
I think the major reason to support escaping the escape character is because people would expect that to work. I tried putting `\\` in a string and it came out as `\`, so maybe it already works? Either way, I think the behavior should be defined and tested. You should be able to use the escape character in the string, definitely by escaping it, and if you don't have to escape it, that's cool too, I guess, although it's a bit weird.

Pod/Classes/BONTag_Private.h
ZevEisenberg: Start and end tags will be stripped from the string. What will happen to escape sequences? Will they be stripped, or won't they?

Pod/Classes/BONText.m
ZevEisenberg: Add a hyphen near the end of the comment

README.md
ZevEisenberg: How about:
> This allows you to apply styles to substrings of localized strings, whose position, order, and even existence may change from language to language.

Imperiopolis:
> @Imperiopolis is it intentional that you can have different escape characters for each tag in a scenario where you're using complex tags? What's the use case?

Yes. I'm not really sure there is a great use case for this, but since these are _complex_ options, I figured it'd be nice to expose it if someone needed it.

ZevEisenberg:
Are there tests for cases with different escapes for different tags?

Imperiopolis:
> Are there tests for cases with different escapes for different tags?

There are now!

Pod/Classes/BONTag.h
Imperiopolis: Discussed with Zev – the only way I know to do this is to mark these functions as `NS_SWIFT_UNAVAILABLE` and then provide swift code to redefine the functions with default parameters.

I'm hesitant to do this as for people who are using the library in a swift-free app, their app will now need to bundle the swift standard library. Could potentially break this into a swift specific subspec, but it feels pretty messy.

Zev is going to investigate if there are any other ways to tackle this.

ZevEisenberg: I was thinking of `NS_SWIFT_NAME`, which lets you describe the function signature, but doesn't appear to let you control things like default parameters. Swift subspec would be nice, but more likely I'll just wait, and rewrite BonMot in Swift for 4.0 or 5.0, once including a Swift dependency in a project is no longer a hardship.

ZevEisenberg: …but here's an idea, @Imperiopolis. You could mark the convenience functions for `BONTagComplexMake` as `NS_SWIFT_UNAVAILABLE`, and add a few more `-initWith...` methods to cover the cases where you would normally use default arguments. You would just never call those in Obj-C code because you would use the convenience functions. Then, the only way to initialize `BONTag` objects in Swift would be to use one of those initializers from Obj-C.

Since it would be a breaking change for Swift users if we marked the convenience functions as unavailable, I'd prefer to do it now. @Imperiopolis, let me know if you'd like to do it, and if not, I'll take care of it.

ZevEisenberg: The other thing I was thinking of was apparently `NS_REFINED_FOR_SWIFT`, but to use it you still have to provide a Swift extension yourself - there's no magic that happens on import from the Obj-C header.

Imperiopolis: `NS_SWIFT_UNAVAILABLE` work is done 👍

PR #151, opened 2016-06-16
Add support for hyphenationFactor

Imperiopolis:
This adds support to `BONText` and `BONChain` for `NSParagraphStyle`'s [`hyphenationFactor` property](https://developer.apple.com/library/ios/documentation/Cocoa/Reference/ApplicationKit/Classes/NSParagraphStyle_Class/#//apple_ref/occ/instp/NSParagraphStyle/hyphenationFactor).

### DIFF
--- a/Pod/Classes/BONChain.h
+++ b/Pod/Classes/BONChain.h
@@ -32,6 +32,7 @@ typedef BONChain *BONCNonnull (^BONChainLineBreakMode)(NSLineBreakMode lineBreak
 typedef BONChain *BONCNonnull (^BONChainParagraphSpacingAfter)(CGFloat paragraphSpacing);
 typedef BONChain *BONCNonnull (^BONChainParagraphSpacingBefore)(CGFloat paragraphSpacingBefore);
 typedef BONChain *BONCNonnull (^BONChainBaselineOffset)(CGFloat baselineOffset);
+typedef BONChain *BONCNonnull (^BONChainHyphenationFactor)(CGFloat hyphenationFactor);
 typedef BONChain *BONCNonnull (^BONChainAlignment)(NSTextAlignment alignment);
 typedef BONChain *BONCNonnull (^BONChainFigureCase)(BONFigureCase figureCase);
 typedef BONChain *BONCNonnull (^BONChainFigureSpacing)(BONFigureSpacing figureSpacing);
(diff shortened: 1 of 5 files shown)


Pod/Classes/BONChain.h
```diff
@@ -78,6 +79,11 @@ typedef BONChain *BONCNonnull (^BONTagComplexStyles)(BONGeneric(NSArray, BONTag
 
 @property (copy, nonatomic, readonly) BONChainBaselineOffset baselineOffset;
 
+/**
+ *  Hyphenation is attempted when the ratio of the text width (as broken without hyphenation) to the width of the line fragment is less than the hyphenation factor. When the paragraph’s hyphenation factor is 0.0, the layout manager’s hyphenation factor is used instead. When both are 0.0, hyphenation is disabled.
```
ZevEisenberg: mmmm, curly apostrophes in comments 👌

ZevEisenberg: It would be nice to have some guidance on what kinds of numbers to use here. I could say the same for Apple's documentation, but if you have any rules of thumb, that would be helpful.

ZevEisenberg:
@Imperiopolis a unit test or two would be nice 😉

Pod/Classes/BONChain.h
Imperiopolis: As far a I understand, it's basically a percentage. 1 means it will 100% use hyphens and 0 means it will never use them.

Imperiopolis:
> @Imperiopolis a unit test or two would be nice 😉

It seemed like the other similar `NSParagraphStyle` properties weren't being tested so I didn't add tests here – can add them if you feel it's important.

ZevEisenberg:
Haha, oops. If you want extra credit, tests for those things would be great. You can enable code coverage in Xcode for the example projects scheme if you want to get an idea of what needs to be tested. Otherwise, please open an issue that the paragraph style stuff should all be tested.

ZevEisenberg:
@Imperiopolis I made a little sample app to test this out and confirm everything is working: http://cl.ly/gXvY. You can see the effects of hyphenation factor around the 0.7-0.9 range if you drag the slider. It's pretty cool, and I learned a thing 😄. After playing with it and rereading the documentation, I understand what you mean about the percentage.

Given that, it would be nice to offer a little guidance as to how this feature works, in addition to just copying the documentation. Specifically, we should mention that only values between 0 and 1 are valid, and that a higher value will result in more aggressive (i.e. more frequent) hyphenation. I had to do some thinking to parse the documentation such that I could understand it, and I'd like to save future users that anguish 😌

ZevEisenberg:
Closing in favor of #156

PR #161, opened 2016-07-11
Chain Concatenation with BONTextables

### LINKED ISSUE #160: Separator Type Inconsistency in Chain Concatenation
Concatenating `BONChain` with `BONText.joinTextables` takes a `BONChain` as a separator, while using `BONChain().appendLink` takes a String as a separator. Should deprecate current version of `appendLink` with `separator` parameter and introduce one with a `separatorTextable` of type `id<BONTextable>`.

nevillco:
- Deprecated appendLink with separator string
- Implemented appendLink with BONTextable
- Updated example code accordingly
- Updated readme accordingly

Fixes #160


Pod/Classes/BONChain.m
```diff
@@ -389,17 +389,18 @@ - (BONTagComplexStyles)tagComplexStyles
...
+- (void)appendLink:(id<BONTextable>)link separatorTextable:(id<BONTextable>)separator
 {
-    if (separator.length > 0) {
+    if (separator.text.string.length > 0) {
         // Recursion!
-        [self appendLink:self.string(separator)]; // add the sparator, with the same properties as self, to the end of the chain.
+        [self appendLink:separator]; // add the sparator, with the same properties as self, to the end of the chain.
     }
-
+
```
ZevEisenberg: #req whitespace. Do you have `clang-format` installed as per the readme? It should catch this.

nevillco: Huh, I thought I did. I'll see what's wrong.

Pod/Classes/BONChain.h
```diff
@@ -119,14 +119,16 @@ typedef BONChain *BONCNonnull (^BONTagComplexStyles)(BONGeneric(NSArray, BONTag
...
-- (void)appendLink:(id<BONTextable>)link separator:(BONNullable NSString *)separator;
+- (void)appendLink:(id<BONTextable>)link separatorTextable:(BONNullable id<BONTextable>)separator;
 
 @end
 
 @interface BONChain (Deprecated)
 
 @property (copy, nonatomic, readonly) BONChainColor textColor __attribute__((deprecated("use -color instead")));
 
+- (void)appendLink:(id<BONTextable>)link separator:(BONNullable NSString *)separator __attribute__((deprecated("use -appendLink:link separatorTextable:separator instead")));
```
ZevEisenberg: #req the way to spell the name of the new method is `-appendLink:separatorTextable:`.

Pod/Classes/BONChain.m
```diff
@@ -389,14 +389,14 @@ - (BONTagComplexStyles)tagComplexStyles
...
 {
-    [self appendLink:link separator:nil];
+    [self appendLink:link separatorTextable:nil];
 }
 
-- (void)appendLink:(id<BONTextable>)link separator:(NSString *)separator
+- (void)appendLink:(id<BONTextable>)link separatorTextable:(id<BONTextable>)separator
 {
-    if (separator.length > 0) {
+    if (separator.text.string.length > 0) {
```
ZevEisenberg: There's an `isEmpty` method (or at least, the name is something like that), which you should use instead, because this `text`'s string could be empty, but its `nextText` might not be, and that method takes care of it.

ZevEisenberg:
#nth this all came about because the documentation and the code didn't agree with each other. Was this code tested before you changed it? If so, please add new tests for the new behavior. If not, some tests would be nice to confirm that the new behavior works consistently.

Pod/Classes/BONChain.h
```diff
@@ -119,14 +119,16 @@ typedef BONChain *BONCNonnull (^BONTagComplexStyles)(BONGeneric(NSArray, BONTag
...
-- (void)appendLink:(id<BONTextable>)link separator:(BONNullable NSString *)separator;
+- (void)appendLink:(id<BONTextable>)link separatorTextable:(BONNullable id<BONTextable>)separator;
 
 @end
 
 @interface BONChain (Deprecated)
 
 @property (copy, nonatomic, readonly) BONChainColor textColor __attribute__((deprecated("use -color instead")));
 
+- (void)appendLink:(id<BONTextable>)link separator:(BONNullable NSString *)separator __attribute__((deprecated("use -appendLink: separatorTextable: instead")));
```
ZevEisenberg: #req no space: `-appendLink:separatorTextable:`. That wasn't a typo - it's just a quirk of how Obj-C selector names are spelled.

Pod/Classes/BONChain.m
```diff
@@ -441,4 +441,14 @@ - (BONChainColor)textColor
     return self.color;
 }
 
+- (void)appendLink:(id<BONTextable>)link separator:(NSString *)separator
+{
+    if (separator.length > 0) {
```
ZevEisenberg: #pp I think you can just forward this to `appendLink:separatorTextable`, and wrap `separator` in a `BONChain`. Might be problematic because you modified the tests instead of adding new tests, so they don't test this method any more. Not sure if it's worth keeping the old and new tests, though ¯\\\_(ツ)\_/¯

Pod/Classes/BONChain.m
```diff
@@ -441,4 +441,9 @@ - (BONChainColor)textColor
     return self.color;
 }
 
+- (void)appendLink:(id<BONTextable>)link separator:(NSString *)separator
+{
+    [self appendLink:link separatorTextable:BONChain.new.string(separator)];
```
ZevEisenberg: #req I may have misled you before. I believe, to keep the original behavior, you need to use `self.string(separator)` here. If you change it, you'll see that some of your test changes result in failing tests.

PR #164, opened 2016-07-15
Left-to-right Debug Strings

### LINKED ISSUE #84: Confusing right-to-left characters
untitled

In the `description` of a `BONText`, right-to-left characters appear, confusingly, _after_ their character names. This should be fixed.

nevillco:
Fixes #84

### DIFF
--- a/Pod/Classes/BONText.m
+++ b/Pod/Classes/BONText.m
@@ -587,6 +587,9 @@ static inline BOOL BONDoublesCloseEnough(CGFloat float1, CGFloat float2)
         if (substringRange.location != 0) {
             [debugString appendString:@"\n"];
         }
+        
+        //Append left-to-right override special character to start of each line
+        [debugString appendString:@"\u202D"];
 
         if ([substring isEqualToString:BONSpecial.objectReplacementCharacter]) {
             BONStringDict *attributes = [originalAttributedString attributesAtIndex:substringRange.location effectiveRange:NULL];
--- a/Pod/Classes/BONText.m
+++ b/Pod/Classes/BONText.m
@@ -578,7 +578,7 @@ static inline BOOL BONDoublesCloseEnough(CGFloat float1, CGFloat float2)
 - (NSString *)debugStringLeftToRight
 {
     NSString *debugString = [self debugString];
-    NSArray<NSString *> *lines = [debugString componentsSeparatedByString:@"\n"];
+    BONGeneric(NSArray, NSString*) *lines = [debugString componentsSeparatedByString:@"\n"];
     NSString *separator = [NSString stringWithFormat:@"\n%@", BONSpecial.leftToRightOverride];
     return [BONSpecial.leftToRightOverride stringByAppendingString:
                                                [lines componentsJoinedByString:separator]];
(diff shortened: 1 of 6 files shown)


Pod/Classes/BONText.m
```diff
@@ -575,6 +575,15 @@ - (NSString *)debugString
     return [self debugStringIncludeImageAddresses:YES];
 }
 
+- (NSString *)debugStringLeftToRight
+{
+    NSString *debugString = [self debugString];
+    NSArray<NSString *> *lines = [debugString componentsSeparatedByString:@"\n"];
```
ZevEisenberg: #req use the `BON*` compatibility macros for generics

scripts/BONSpecialGenerator.swift
```diff
@@ -66,6 +67,7 @@ let charactersRequiringFormatStrings: Set<unichar> = [
 // These special characters are to be excluded from the human readable string dictionary.
 let charactersToExcludeFromHumanReadableStringDictionary: Set<unichar> = [
     0x0020,
+    0x202D
```
ZevEisenberg: #? Do we really want to exclude this from human readable strings? If this character is present in a string I'm debugging, I feel like I'd want to know about it.

nevillco: Yeah you're right. Not sure why I originally decided on that.

scripts/BONSpecialGenerator.swift
```diff
@@ -65,7 +66,7 @@ let charactersRequiringFormatStrings: Set<unichar> = [
 
 // These special characters are to be excluded from the human readable string dictionary.
 let charactersToExcludeFromHumanReadableStringDictionary: Set<unichar> = [
-    0x0020,
+    0x0020
```
ZevEisenberg: #nth don't kill this comma. The comma is there to prevent future additions to this array from messing with the revision history of previous lines. Kind of a moot point now, unless you rebase, but a good habit to get into.

ZevEisenberg:
#req please port this change to `BONChain` as well. It's still calling `[self.text debugStringIncludeImageAddresses:YES]` to get the debug string.

Pod/Classes/BONText.m
```diff
@@ -575,6 +575,15 @@ - (NSString *)debugString
     return [self debugStringIncludeImageAddresses:YES];
 }
 
+- (NSString *)debugStringLeftToRight
```
ZevEisenberg: #nth thinking this should be able to pass through image addresses, in case you need to use it for a unit test. Or would you always just call `-debugStringIncludeImageAddresses:NO` in that case? What do you think, @nevillco? Would the output of this method ever need to be deterministic?

nevillco: I was on the fence about that; I suppose it would be nice to have for someone that wants to use it in unit tests. Only tradeoff is naming (ie. `debugStringLeftToRightIncludeImageAddresses`) could be ugly. Was also considering some sort of refactoring to use a single `debugString` with parameters for both `leftToRight` and `includeImageAddresses`.

ZevEisenberg: I'm thinking that, for a debug string, the LTR character is more cruft that you don't want in your in-code debug strings, so it's probably OK to leave it as-is.

PR #185, opened 2016-09-22
Update to Swift 3

Anders123fr:
Updated to Swift 3 using the migrator in Xcode 8

### DIFF
--- a/Example/BonMot.xcodeproj/project.pbxproj
+++ b/Example/BonMot.xcodeproj/project.pbxproj
@@ -302,6 +302,9 @@
 					6003F589195388D20070C39A = {
 						LastSwiftMigration = 0800;
 					};
+					6003F5AD195388D20070C39A = {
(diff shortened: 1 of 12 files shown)


KingOfBrian:
Hey @Anders123fr -- thanks for the PR, but it cuts a bit too deep. The 3.0 converter actually ran on all the swift 2.3 compatibility code. I'll push up a commit tonight, but I'm guessing Xcode just prompted the converter because I hadn't opened with the GM yet.

Also, just a warning that if you are using this branch, the API is still in flux. We'd love any API feedback on the other issues here.

Thanks!

Brian

KingOfBrian:
Actually Anders, I didn't get any converter prompt. Were you having an issue building with Swift 3? It should be working fine.

Anders123fr:
Okay I just couldn't compile it, so i tried to fork it and i was prompted with the converter.

Issue #174, opened 2016-09-06
Restore the BONTextConstraint


KingOfBrian:
- [ ] Restore the Objective-C BONTextConstraint. 
- [ ] Restore the constraint tests.
- [ ] Restore example app
- [ ] Optionally re-write it in swift.

noremac:
Merged!

PR #202, opened 2016-10-04
Add TextAlignmentConstraint for https://github.com/Raizlabs/BonMot/issues/174

noremac:


Sources/UIKit/TextAlignmentConstraint.swift
```diff
@@ -0,0 +1,162 @@
...
+//
+
+#if os(OSX)
+    import AppKit
+#else
+    import UIKit
+#endif
+
+@objc(BONTextAlignmentConstraintAttribute)
+public enum TextAlignmentConstraintAttribute: Int {
```
ZevEisenberg: Would prefer to have this nested inside `TextAlignmentConstraint`, so it's `BonMot.TextAlignmentConstraint.Attribute`. The Obj-C name is good, though.

Sources/UIKit/TextAlignmentConstraint.swift
```diff
@@ -0,0 +1,162 @@
...
+            return .top
+        case .lastBaseline, .bottom:
+            return .bottom
+        case .unspecified:
+            return .notAnAttribute
+        }
+    }
+}
+
+private var TextAlignmentConstraintKVOContext = "TextAlignmentConstraintKVOContext" as NSString
```
ZevEisenberg: Might want to add BonMot to the string, to make debugging easier and provide even better defense against collisions. Also, might want to embed this inside `TextAlignmentConstraint`

Sources/UIKit/TextAlignmentConstraint.swift
```diff
@@ -0,0 +1,162 @@
...
+
+@objc(BONTextAlignmentConstraint)
+public class TextAlignmentConstraint: NSLayoutConstraint {
+
+    public private(set) var firstItemAttribute: TextAlignmentConstraintAttribute = .unspecified
+    public private(set) var secondItemAttribute: TextAlignmentConstraintAttribute = .unspecified
+    private var item1: AnyObject!
+    private var item2: AnyObject!
+
+    public static func with(item view1: AnyObject, attribute attr1: TextAlignmentConstraintAttribute, relatedBy relation: NSLayoutRelation, toItem view2: AnyObject, attribute attr2: TextAlignmentConstraintAttribute) -> TextAlignmentConstraint {
```
ZevEisenberg: Can we make this an initializer?

Sources/UIKit/TextAlignmentConstraint.swift
```diff
@@ -0,0 +1,162 @@
...
+
+    private func tearDownObservers() {
+        for keyPath in fontKeyPaths {
+            removeObserver(self, forKeyPath: keyPath, context: &TextAlignmentConstraintKVOContext)
+        }
+    }
+
+    private var fontKeyPaths: [String] {
+        let firstItemFontKeyPath = [
+            #selector(getter: NSLayoutConstraint.firstItem),
```
ZevEisenberg: I think this could be simplified as `"\(#selector(getter: NSLayoutConstraint.firstItem).\(FontSelector)"`.

Sources/UIKit/TextAlignmentConstraint.swift
```diff
@@ -0,0 +1,162 @@
...
+            font = item.perform(FontSelector).takeUnretainedValue() as? BONFont
+        }
+
+        return font
+    }
+}
+
+// The class part of this selector is ignored; it is there simply to satisfy Xcode's new selector syntax.
+#if os(OSX)
+    private let FontSelector = #selector(getter: NSTextField.font)
```
ZevEisenberg: I'd prefer to just construct these from string literals (`"font"`). I think you can do that, right?

Sources/UIKit/TextAlignmentConstraint.swift
```diff
@@ -0,0 +1,162 @@
...
+            ].map({ $0.description })
+            .joined(separator: ".")
+
+        let secondItemFontKeyPath = [
+            #selector(getter: NSLayoutConstraint.secondItem),
+            FontSelector
+            ].map({ $0.description })
+            .joined(separator: ".")
+
+        return [firstItemFontKeyPath, secondItemFontKeyPath]
```
ZevEisenberg: I'm thinking we should also observe `attributedText`, if the object exposes it. If you set the `attributedText` on a `UILabel`, I think it will change `font` without triggering KVO for it.

Sources/UIKit/TextAlignmentConstraint.swift
noremac: When constructing strings that'd be fine, but that usage is deprecated when using `response(to: `.

ZevEisenberg: You are correct. Boo. Good comment, then. I'd nest it as a `static let` inside the constraint subclass, and use a lowercase initial letter.

Sources/UIKit/TextAlignmentConstraint.swift
```diff
@@ -0,0 +1,268 @@
...
+        case firstBaseline
+        case lastBaseline
+        case bottom
+
+        var layoutAttribute: NSLayoutAttribute {
+            switch self {
+            case .top, .capHeight, .firstBaseline, .xHeight:
+                #if swift(>=3.0)
+                    return .top
+                #else
```
KingOfBrian: You should add an extension on `NSLayoutAttribute` in `Compatibility.swift` and add:
```
extension NSLayoutAttribute {
   static var top = NSLayoutAttribute.Top
}
```
Ideally, once you've written the 3.0 code, getting 2.3 to work should be adding stuff to `Compatibility.swift`.

Sources/UIKit/TextAlignmentConstraint.swift
```diff
@@ -0,0 +1,268 @@
...
+        #endif
+    #else
+        #if swift(>=3.0)
+        private static let fontSelector = #selector(getter: UITextField.font)
+        private static let attributedTextSelector = #selector(getter: UITextField.attributedText)
+        #else
+        private static let fontSelector = Selector("font")
+        private static let attributedTextSelector = Selector("attributedText")
+        #endif
+    #endif
```
KingOfBrian: This is an interesting one. You should be able to to drop the OS difference by using a typealias and using BONTextField (which will map to NS or UI). Then there's just 1 #if conditional

Tests/TextAlignmentConstraintTests.swift
```diff
@@ -0,0 +1,223 @@
...
+@testable import BonMot
+
+class TextAlignmentConstraintTests: XCTestCase {
+
+    private func field(withText text: String, fontSize: CGFloat) -> BONTextField {
+        let field = BONTextField(frame: .zero)
+        field.translatesAutoresizingMaskIntoConstraints = false
+
+        #if swift(>=3.0)
+            field.font = .systemFont(ofSize: fontSize)
```
ZevEisenberg: These could break if the system font metrics change. Not sure if it's worth it now, but it might be nice to change these to a hardcoded font.

PR #206, opened 2016-10-07
[4.0] API Exploration

KingOfBrian:
This PR is an exploration of adding an object similar to `BONChain` into the 4.0 API. Also, if you open [`Example/BonMot-Example-iOS/DemoStrings.swift`](https://github.com/Raizlabs/BonMot/compare/feature/swift-first-XML-Style...feature/swift-first-chain-exploration?expand=1#diff-2ca1e5052419ffa5225a1ad9d0f6b4e4R1), I split up the demo strings to show the composition of the styles rather than one huge method call with nested method calls. I think that was masking how the API could be composed.

Also of note, is in the new version of swift, We'll be able to make an Array of `Composable`s conform to `Composable`, which will be very nice I believe


ZevEisenberg:
I asked around, and this exploration, though informative, turned out to be a dead end.

PR #208, opened 2016-10-07
Naming things

ZevEisenberg:
It was all @KingOfBrian's idea.

### DIFF
--- a/Tests/AttributedStringStyleTests.swift
+++ b/Tests/AttributedStringStyleTests.swift
@@ -153,7 +153,7 @@ class AttributedStringStyleTests: XCTestCase {
             .paragraphSpacingBefore(1),
             .hyphenationFactor(1)
             )
-        style.update(attributedStringStyle: AttributedStringStyle.style(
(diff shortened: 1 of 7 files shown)


Tests/AttributedStringStyleTests.swift
```diff
@@ -215,17 +215,17 @@ class AttributedStringStyleTests: XCTestCase {
     //   - an empty style object that is updated by the passed style object
     //   - a fully populated style object that is updated by the passed style object
     func checks(for style: AttributedStringStyle) -> [(style: AttributedStringStyle, fullStyle: Bool)] {
-        let derived = AttributedStringStyle.style().derive { derived in
+        let derivedStyle = AttributedStringStyle.style().byAdding { derivedStyle in
             // Not sure why but if this line is commented out, the line after is a compilation error.
```
ZevEisenberg: To what? It's still a derived style. Or do you want to just strip out the block and do this one inline?

KingOfBrian:
You were the one complaining about `derive` -- I just agreed.

PR #226, opened 2016-11-01
Add currentStyle to XMLStyler protocol's style function.

Imperiopolis:
As discussed in slack, this allows for use cases such as follows to support custom styles based on the parent style:
```swift
func style(forElement name: String, attributes: [String: String], currentStyle: StringStyle) -> StringStyle? {
    let currentFont = currentStyle.font ?? baseFont
    switch name.lowercased() {
    case "i":
        fallthrough
    case "em":
        guard currentFont.fontName != FontName.verlagBold.stringValue else {
            return StringStyle.style(Styles.font(.verlagBoldItalic, size: currentFont.pointSize))
        }

        return StringStyle.style(Styles.font(.verlagBookItalic, size: currentFont.pointSize))
    case "b":
        fallthrough
    case "strong":
        guard currentFont.fontName != FontName.verlagBookItalic.stringValue else {
            return StringStyle.style(Styles.font(.verlagBoldItalic, size: currentFont.pointSize))
        }

        return StringStyle.style(Styles.font(.verlagBold, size: currentFont.pointSize))
    case "font":
        guard let sizeString = attributes["size"],
            let size = NumberFormatter().number(from: sizeString) as? CGFloat else {
                fallthrough
        }
        return StringStyle.style(.font(currentFont.withSize(size)))
    default:
        return nil
    }
}
```

### DIFF
--- a/Sources/XMLBuilder.swift
+++ b/Sources/XMLBuilder.swift
@@ -95,8 +95,8 @@ public enum XMLStyleRule {
 
 /// This contract is used to transform an XML string into an attributed string.
 public protocol XMLStyler {
-    /// Return the style to apply for to the contents of the element. The style is sadded onto the current style
-    func style(forElement name: String, attributes: [String: String]) -> StringStyle?
+    /// Return the style to apply for to the contents of the element. The style is added onto the current style
+    func style(forElement name: String, attributes: [String: String], currentStyle: StringStyle) -> StringStyle?
 
     /// Return a string to extend into the string being built. This is done after the style for the element has been applied, but before the contents of the element.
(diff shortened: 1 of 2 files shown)


Sources/XMLBuilder.swift
```diff
@@ -95,8 +95,8 @@ public enum XMLStyleRule {
 
 /// This contract is used to transform an XML string into an attributed string.
 public protocol XMLStyler {
-    /// Return the style to apply for to the contents of the element. The style is sadded onto the current style
```
ZevEisenberg: Joyable employees really pull their weight! ;)

PR #228, opened 2016-11-02
Add [T] variants for all functions that take type T...

Imperiopolis:


Imperiopolis:
It'd be ideal if someday `[T]` and `T...` were interchangeable when it's the last parameter, but for now it seems to make sense to support either variety. My particular need for this is something like this:

```swift
func style(_ name: StyleType, _ otherStyles: StringStylePart...) -> StringStyle {
    return style(name, otherStyles)
}

func style(_ name: StyleType, _ otherStyles: [StringStylePart] = []) -> StringStyle {
    return name.style.byAdding(otherStyles)
}
```

KingOfBrian:
Another method of using the existing API would be:

```
name.style.byAdding(style: StringStyle(parts: otherStyles))
```

Also, curious as to what StyleType is!

Imperiopolis:
True, that would work and happy to do it that way if you think it makes more sense than supporting `[T]` and `T...`

`StyleType` is an enum with things like `h1`, `h2`, `body`, `caption`, etc. that sets up some base styles. Generally these are font face, size, weight, and occasionally a few other things like number spacing / casing

Imperiopolis:
Thinking about it more, the way I'd ideally love to construct this is with another type on `StringStylePart` that takes a `StringStyle` so I could do something like...
```swift
extension StringStylePart {
    static var h2: StringStylePart {
        return .combined(.font(...), .color(.black), .lineHeight(...))
    }
}
```

and then use it like 

```swift
StringStyle.style(.h2, , .alignment(.center), .color(.blue))
```

Sources/StylePart.swift
```diff
@@ -76,31 +83,25 @@ extension StringStyle {
     }
     #endif
 
-    #if swift(>=3.0)
```
ZevEisenberg: Including the "AllTheThings" target?

KingOfBrian:
So I like the call site usage there and it should be pretty easy to support. I'd prefer to have a case `.style(StringStyle)`, than `.combined([StylePart])`, but I think either would work.

I wrote a long ditty about style and style part not mixing before getting what you were saying. But I came up with an API which may also be worth looking into as well:

```swift
    public func styled(with style: StringStyle, _ overrideParts: StyleParts...) -> NSAttributedString {...}
```

Which will allow a similar, but more constrained call site usage. Let's let the thought cook over the weekend and pick one or both of them.

Very glad to have an advanced user of the API already, it's a huge help!

ZevEisenberg:
I think `.style(StringStyle)` is better, since this is a fairly advanced feature, so best to support the version that offers the most flexibility, e.g. combining multiple different styles.

Imperiopolis:
Agree with Zev that `.style(StringStyle)` is better, but I'd definitely also be okay with both. 

I'll implement them and submit a new PR and kill this one, and come monday we can decide if one of the versions should be dropped :)

PR #230, opened 2016-11-02
CircleCI migration

dostrander:
- it works… Not sure if we are 100% sold on the transfer but here it is either way.


circle.yml
```diff
@@ -0,0 +1,21 @@
...
+  pre:
+    - brew update 1> /dev/null 2> /dev/null
+    - brew outdated carthage || brew upgrade carthage
+
+checkout:
+  post:
+    - git submodule update --init --recursive
+test:
+  override:
+    - ./scripts/macOS
```
ZevEisenberg: #req these scripts should have a file extension

circle.yml
```diff
@@ -0,0 +1,21 @@
...
+    version: "8.0"
+
+dependencies:
+  pre:
+    - brew update 1> /dev/null 2> /dev/null
+    - brew outdated carthage || brew upgrade carthage
+
+checkout:
+  post:
+    - git submodule update --init --recursive
```
ZevEisenberg: We probably don't need to do this

circle.yml
```diff
@@ -0,0 +1,21 @@
+machine:
+  xcode:
+    version: "8.0"
+
+dependencies:
+  pre:
```
ZevEisenberg: #req I believe we need a `bundle install` and, depending on how Circle is set up, a `gem install bundler`

circle.yml
ZevEisenberg: #nth since the build scripts are all so similar, they could be refactored to be the same script four times with passed variables for project, scheme, sdk, and destination.

circle.yml
dostrander: It infers the bundle install step

circle.yml
dostrander: Yeah I was going to do that but then I got lazy but I think I'm going to again today. Might just end up using fast lane because it's already installed on the machine

ZevEisenberg: #pp I think the build script is simple enough that we don't need to introduce a dependency on fastlane.

ZevEisenberg:
In addition to my review comments, the Circle build appears to be failing. Also, this branch should delete `.travis.yml`.

ZevEisenberg:
@dostrander also, please replace the Travis status badge in the readme with a Circle status badge.

circle.yml
dostrander: @ZevEisenberg It's slightly too complex to do in shell alone and i don't want to mess with a ruby environment right now. I'm going to add "clean up build scripts" as an issue so it doesn't get lost and add the `.sh` extensions.

scripts/iOS.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+
+set -o pipefail && \
+  xcodebuild clean build test \
+  -project BonMot.xcodeproj \
+  -scheme BonMot-iOS \
+  -sdk iphonesimulator \
+  -destination "id=E8DD285C-51EE-4DB5-B326-7E927686EC36" \
```
ZevEisenberg: #? why IDs? It seems like this will be super fragile - every time the simulators are regenerated, the IDs will be different. I don't think they're deterministic, are they? What was wrong with using names like on Travis?

PR #231, opened 2016-11-03
173 opentype features

### LINKED ISSUE #173: OpenType Feature Support
Provide better feature support for specifying font features and checking to see if a font supports a given feature.
- [x] superscript
- [x] subscript
- [x] small caps
  - you can enable `kLowerCaseType` and `kUpperCaseType` on at the same time
- [x] ligatures
- [x] stylistic alternates
- ~contextual alternates~ too niche for 4.0 release
- [x] tests
- [ ] example app
- [x] nice-to-have: query OpenType feature support: #35

ZevEisenberg:
Closes #173.


Sources/StylePart.swift
```diff
@@ -174,8 +191,33 @@ extension StringStyle {
...
+                    self.fontFeatureProviders += [smallCaps as FontFeatureProvider]
+                    return
+                }
+                else if case let .stylisticAlternates(stylisticAlternates) = stylePart {
+                    self.fontFeatureProviders += [stylisticAlternates as FontFeatureProvider]
+                    return
+                }
                 else if case let .fontFeature(featureProvider) = stylePart {
                     self.fontFeatureProviders.append(featureProvider)
+                    return
```
KingOfBrian: 1) Why is there a return? I see it was already there, but it doesn't seem needed
2) The StringStyle object models these features in properties. They should be transferred into those properties, not into the fontFeatureProviders array. That array was sort of a generic bucket, I think we may way to look at it again.

ZevEisenberg:
If git-bisect is to be believed, https://github.com/Raizlabs/BonMot/pull/231/commits/878e8e73d7b0c9f61afcbc514a76a75be5365941 is the commit that broke this build. 
tumblr_inline_oaefr0iq2w1qjio1o_500

ZevEisenberg:
Followup: it looks like something in cd3f8ff0ff5526b9ce9394064157f3a864655a5b caused the way fonts are composed to break `==` for `NSFont` on macOS…sometimes. Fixed it by comparing font descriptors instead of fonts (https://github.com/Raizlabs/BonMot/pull/231/commits/c9c436987a264f62613c736e7ed225e3c4d91a2a), because apparently, `==` works for font descriptors on all platforms in case where regular font equality checking fails.

Sources/StringStyle.swift
```diff
@@ -93,7 +107,19 @@ extension StringStyle {
         #if os(iOS) || os(tvOS) || os(OSX)
             // Apply the features to the font present
             let preFeaturedFont = theAttributes[NSFontAttributeName] as? BONFont
-            let featuredFont = preFeaturedFont?.font(withFeatures: fontFeatureProviders)
+            var featureProviders = fontFeatureProviders
+
+            featureProviders += numberCase.map { $0 as FontFeatureProvider }.asArray
```
ZevEisenberg: It needs to be `flatMap`, but yeah, that's a nice solution.

PR #233, opened 2016-11-04
Adds `.styled(StringStyle)` part and `styled(with:overrideParts:)`

Imperiopolis:
Based on conversation here: https://github.com/Raizlabs/BonMot/pull/228

I ended up still having to implement a `[T]` version `StringStyle(_:StylePart...)` in order to be able to create `styled(with:overrideParts:)`, but let me know if there is another way to do this I'm missing.

### DIFF
--- a/Tests/AttributedStringStyleTests.swift
+++ b/Tests/AttributedStringStyleTests.swift
@@ -272,4 +272,15 @@ class StringStyleTests: XCTestCase {
         return [(style: style, fullStyle: false), (style: emptyStyle, fullStyle: false), (style: updated, fullStyle: true)]
     }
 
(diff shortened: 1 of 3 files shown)


Tests/AttributedStringStyleTests.swift
```diff
@@ -272,4 +272,15 @@ class StringStyleTests: XCTestCase {
         return [(style: style, fullStyle: false), (style: emptyStyle, fullStyle: false), (style: updated, fullStyle: true)]
     }
 
+    func testStyleStylePart() {
```
ZevEisenberg: I suppose the name makes sense, but `StyleStylePart` looks confusing.

PR #237, opened 2016-11-07
Re-scope StylePart as StringStyle.Part

### LINKED ISSUE #232: Change StylePart to StringStyle.Part
Instead of having `StylePart` as a top-level enum, nest it in an extension of `StringStyle`:

```swift
extension StringStyle {

    public enum Part {
        ...
    }

}
```

What do people think? Comment below!

ZevEisenberg:
Fixes #232. Requires #231 to be merged first.


Sources/StringStyle+Part.swift
```diff
@@ -11,60 +11,65 @@
...
+        case textStyle(BonMotTextStyle)
+        #endif
+        #if os(iOS) || os(tvOS)
+        case adapt(AdaptiveStyle)
+        #endif
+
+    }
 
     // An advanced part that allows combining multiple parts as a single part
     case style(StringStyle)
```
Imperiopolis: Looks like a bad merge? This should be in the enum.

ZevEisenberg: Thought I fixed that. Taking a look. Yes, it was a bad merge.

Imperiopolis:
Why is `StringStyle+Part` broken into three extensions? It feels very unclear to me what the distinction is between these extensions – perhaps add comments delineating what each one is for to make any future additions clear?

ZevEisenberg:
@Imperiopolis I merged the extensions, since there wasn't much point in having them. Should be good to go now.

PR #245, opened 2016-11-10
Speed up linting

ZevEisenberg:
Thanks to @ateliercw for the wonderful hack


ZevEisenberg:
Linting was taking about 8 minutes. With this workaround, it's now under 3 minutes. @ateliercw++ 🎉

Issue #155, opened 2016-07-05
Add support for Speech Attributes for Attributed Strings


ZevEisenberg:
via https://twitter.com/Sommer/status/750432069214351361: https://developer.apple.com/library/tvos/documentation/UIKit/Reference/UIAccessibility_Protocol/index.html#//apple_ref/doc/constant_group/Speech_Attributes_for_Attributed_Strings

PR #263, opened 2017-01-24
Accessibility

aral:
The demo strings in the iOS Examples app are currently not accessible.

Specifically:

- The XML example has a decorative image that is visible to the accessibility system and shouldn’t be.
- The Composition example has a content image without alternative text.
- The Images & Special characters example has content images without alternative text.

Furthermore, there is a bug in the `tintedImage(color:)` method in the `BONImage` extension that loses the accessibility label on the original image.

This pull request:

1. Fixes the `tintedImage(color:)` method in the `BONImage` extension so that the resulting tinted image has the same accessibility label as the original image.
2. Adds accessibility support to the images in the iOS Example app.
3. (Minor) Adds the .idea folder added to projects by AppCode to .gitignore.

Regarding the second point, the difference in experience of a person using the iOS Example app with VoiceOver is the following:

img_5887

### XML example

**Before:** Maria Sharapova attachment png, crisp, very dark, file

**After:** Maria Sharapova

* * *

### Composition:

**Before:** You’re going to need a bigger attachment, png, crisp, dark, file

**After:** You’re going to need a bigger boat

* * *

### Images & Special Characters:

**Before:** 2 2

**After:** 2 bee oar knot 2 bee

### DIFF
--- a/Sources/Image+Tinting.swift
+++ b/Sources/Image+Tinting.swift
@@ -64,6 +64,9 @@ public extension BONImage {
         // Prevent further tinting
         image.isTemplate = false
 
+        // Transfer accessibility label
+        image.accessibilityLabel = self.accessibilityLabel
+
         return image
     }
     #else
@@ -112,6 +115,9 @@ public extension BONImage {
         if !UIEdgeInsetsEqualToEdgeInsets(originalCapInsets, image.capInsets) || originalResizingMode != image.resizingMode {
             image = image.resizableImage(withCapInsets: originalCapInsets, resizingMode: originalResizingMode)
         }
+
+        // Transfer accessibility label
+        image.accessibilityLabel = self.accessibilityLabel
 
         return image
     }
--- a/Sources/Image+Tinting.swift
+++ b/Sources/Image+Tinting.swift
@@ -64,8 +64,8 @@ public extension BONImage {
         // Prevent further tinting
         image.isTemplate = false
 
-        // Transfer accessibility label
-        image.accessibilityLabel = self.accessibilityLabel
+        // Transfer accessibility description
+        image.accessibilityDescription = self.accessibilityDescription
 
         return image
     }
@@ -116,8 +116,10 @@ public extension BONImage {
             image = image.resizableImage(withCapInsets: originalCapInsets, resizingMode: originalResizingMode)
         }
 
-        // Transfer accessibility label
+        // Transfer accessibility label (iOS only; watchOS does not have accessibilityLabel on UIImage).
+        #if os(iOS)
         image.accessibilityLabel = self.accessibilityLabel
+        #endif
 
         return image
     }
--- a/Sources/Image+Tinting.swift
+++ b/Sources/Image+Tinting.swift
@@ -116,9 +116,9 @@ public extension BONImage {
             image = image.resizableImage(withCapInsets: originalCapInsets, resizingMode: originalResizingMode)
         }
 
-        // Transfer accessibility label (iOS only; watchOS does not have accessibilityLabel on UIImage).
-        #if os(iOS)
-        image.accessibilityLabel = self.accessibilityLabel
+        // Transfer accessibility label (watchOS not included; does not have accessibilityLabel on UIImage).
+        #if os(iOS) || os(tvOS)
+            image.accessibilityLabel = self.accessibilityLabel
         #endif
 
         return image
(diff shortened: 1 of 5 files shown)


ZevEisenberg:
@aral it's saying "file", not "foul" (I slowed down the speaking rate to check).

If you make a change to the branch (even a trivial one), I've enabled CircleCI builds for fork pull requests.

ZevEisenberg:
@aral it looks like `accessibilityLabel` on iOS/tvOS maps to `accessibilityDescription` on the Mac. Would you mind updating your branch to fix the Mac build? You can test it locally by running tests in the `BonMot-OSX` target or building the `AllTheThings` target.

ZevEisenberg:
@aral oh, and it also looks like `accessibilityLabel` isn't available on `UIImage` on watchOS, so you'll probably want to conditionally include that line as well.

Tests/ImageTintingTests.swift
```diff
@@ -85,4 +85,13 @@ class ImageTintingTests: XCTestCase {
         BONAssertEqualImages(untintedResult!, tintAttemptResult!)
     }
 
+    func testAccessibilityLabelTransferToTintedImage() {
+        let accessibilityLabel = "I’m the very model of a modern accessible image."
+        let accessibleImage = UIImage(cgImage: imageForTest.cgImage!)
```
aral: Thanks for all the notes, Zev. Will update today. Sorry for missing the cross-platform issues.

ZevEisenberg: No problem! It's pretty rare for a project like this. I should make sure the readme includes enough info on contributing. I never imagined we'd get as much community involvement as we have!

aral: Looked into it on WatchKit a bit and couldn’t come up with anything. I don’t have any Apple Watch apps at the moment. Will look into it again when I can and possibly file a radar. Do you know of anyone using the framework in an open source Watch app I can look at?

aral:
The latest commits should address all the review points. Do let me know if I missed anything.

Sources/Image+Tinting.swift
```diff
@@ -113,6 +116,11 @@ public extension BONImage {
             image = image.resizableImage(withCapInsets: originalCapInsets, resizingMode: originalResizingMode)
         }
 
+        // Transfer accessibility label (iOS only; watchOS does not have accessibilityLabel on UIImage).
+        #if os(iOS)
```
ZevEisenberg: What about tvOS?

Sources/Image+Tinting.swift
```diff
@@ -113,6 +116,11 @@ public extension BONImage {
...
+        image.accessibilityLabel = self.accessibilityLabel
```
ZevEisenberg: nice-to-have: indent this line (project style: anything enclosed in `#if` should be indented.

Tests/ImageTintingTests.swift
```diff
@@ -65,7 +67,7 @@ class ImageTintingTests: XCTestCase {
         XCTAssertNotNil(tintedResult)
 
         BONAssertNotEqualImages(untintedResult!, tintedResult!)
-    }
+    }
```
ZevEisenberg: Extra whitespace. (If you `brew install swiftlint` 0.16.1 or higher, it will catch this for you. You can also enable trimming trailing whitespace in Xcode → Preferences → Text Editing → Automatically trim trailing whitespace.)

Tests/ImageTintingTests.swift
```diff
@@ -84,5 +86,23 @@ class ImageTintingTests: XCTestCase {
 
         BONAssertEqualImages(untintedResult!, tintAttemptResult!)
     }
+    
+    func testAccessibilityIOS() {
+        #if os(iOS)
```
ZevEisenberg: Same as above: what about tvOS?
Also: you could just do a single `testAccessibility` function, which the contents wrapped in `#if`, instead of two separate functions. I don't strongly care either way, I guess, but it might let you share more lines of code between the tests.

Tests/ImageTintingTests.swift
ZevEisenberg: I don't. I haven't seen much watchOS development in the wild.

Sources/Image+Tinting.swift
aral: Done.

Sources/Image+Tinting.swift
aral: Done :)

Tests/ImageTintingTests.swift
aral: swiftlint installed – will go mumble to myself about brace placement now ;)

Tests/ImageTintingTests.swift
aral: I initially had it in one method but decided it made the code harder to read. Updated to include tvOS.

ZevEisenberg:
@aral thanks so much for your contribution! I'm going to try to get a couple more AX things in (https://github.com/Raizlabs/BonMot/issues/155) and cut a new release today.

PR #267, opened 2017-01-31
Overriding styles on attributed strings

ZevEisenberg:
This PR contains a currently-failing test, demonstrating what I believe to be broken behavior. I wanted to run it by @KingOfBrian and others before taking the time to fix it.

- [x] wait for #296 to merge, and then re-target into `master`

### DIFF
--- a/Sources/Composable.swift
+++ b/Sources/Composable.swift
@@ -98,6 +98,19 @@ public extension NSAttributedString {
         return string
     }
 
+    public func styled(with style: StringStyle, _ overrideParts: StringStyle.Part...) -> NSAttributedString {
+        let newStyle = style.byAdding(overrideParts)
+        let newAttributes = newStyle.attributes
+
+        let mutableSelf = mutableStringCopy()
+        let fullRange = NSRange(location: 0, length: mutableSelf.string.utf16.count)
+
+        for (key, value) in newAttributes {
+            mutableSelf.addAttribute(key, value: value, range: fullRange)
+        }
+        return mutableSelf.immutableCopy()
+    }
+
 }
 
 extension NSAttributedString: Composable {
--- a/Sources/MutableCopying.swift
+++ b/Sources/MutableCopying.swift
@@ -23,6 +23,12 @@ extension NSAttributedString {
         return copy
     }
 
+    @nonobjc func immutableCopy() -> NSAttributedString {
+        guard let copy = copy() as? NSAttributedString else {
+            fatalError("Failed to copy() \(self)")
+        }
+        return copy
+    }
 }
 
 extension NSParagraphStyle {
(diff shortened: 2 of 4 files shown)


ZevEisenberg:
I got distracted while debugging this, and ended up filing a Swift issue which makes it hard to chase this stuff: https://bugs.swift.org/browse/SR-3886

ZevEisenberg:
Notes to self while debugging: it looks like the instance method `StringStyle.supplyDefaults(for attributes: StyleAttributes?) -> StyleAttribtues` is either buggy or poorly named (or both), and that it's supplying defaults in the wrong direction from what is expected - that is, the parent is overriding the child, when it should be the other way around.

ZevEisenberg:
@KingOfBrian would you mind having a look at this when you have a chance? After months of on-and-off debugging with more and more complex adjustments, I realized that I could sidestep the entire issue of method naming and figuring out what was actually going on. The fundamental problem is that the default protocol implementation of `Composable`'s `public func styled(with style: StringStyle, _ overrideParts: StringStyle.Part...) -> NSAttributedString` doesn't do what we want it to in the case of `NSAttributedString`, so we just override it in the case of attributed strings and do something different.

Sources/Composable.swift
```diff
@@ -98,6 +98,19 @@ public extension NSAttributedString {
         return string
     }
 
+    public func styled(with style: StringStyle, _ overrideParts: StringStyle.Part...) -> NSAttributedString {
+        let newStyle = style.byAdding(overrideParts)
+        let newAttributes = newStyle.attributes
+
+        let mutableSelf = mutableStringCopy()
```
Imperiopolis: Wondering if this could be simplified to something like this?

```swift
public func styled(with style: StringStyle, _ overrideParts: StringStyle.Part...) -> NSAttributedString {
    return string.styled(with: style, overrideParts)
}
```

And then changing up the `Composable` extension to something like this
```swift
    public func styled(with style: StringStyle, _ overrideParts: StringStyle.Part...) -> NSAttributedString {
        return styled(with: style, overrideParts)
    }

    public func styled(with style: StringStyle, _ overrideParts: [StringStyle.Part]) -> NSAttributedString {
        let string = NSMutableAttributedString()
        let newStyle = style.byAdding(stringStyle: StringStyle(overrideParts))
        append(to: string, baseStyle: newStyle)
        return string
    }
```

Or is there other data on the `NSAttributedString` we're concerned about copying over besides attributes?

ZevEisenberg: This causes the new test introduced in this PR to fail. I don't remember in great detail what's going on in that case, but the method names in this area are kind of borked, and it was an unrelated yak shave to try to clean them up. This method needs to behave fundamentally differently from the others that look like it, in terms of what's being inherited and what's being overridden, but it's hard to articulate why because it takes a while for me just to load the meanings of each of the poorly-named methods into my mental RAM 😩. Definitely good fodder for future cleanup, but I really wanted to get this bug fixed quickly, and the tests give me confidence that, bad naming be damned, this is ready to go.

PR #270, opened 2017-02-19
Add support for kFractionsType

### LINKED ISSUE #243: Add support for kFractionsType
```swift
/*
 *  Summary:
 *    Selectors for feature type kFractionsType
 */

public var kNoFractionsSelector: Int { get }
public var kVerticalFractionsSelector: Int { get }
public var kDiagonalFractionsSelector: Int { get }
```

EthanSchatzline:
Fixes #243 

- added Fractions as a FontFeatureProvider
    - it has 3 cases:
        - disabled: kNoFractionsSelector
        - diagonal: kDiagonalFractionsSelector
        - vertical: kVerticalFractionsSelector
- added a test for fractions
- created a fractions example in DemoStrings
- added the fraction example to the end of the OpenType Features section


Sources/FontFeatures.swift
```diff
@@ -97,6 +97,34 @@
...
+        case none
+
+        /// Diagonal Fractions, when written on paper, are written on one line
+        /// with the numerator diagonally above and to the left of the
+        /// demoninator, with the slash ("/") between them.
+        case diagonal
+
+        /// Vertical Fractions, when written on paper, are written on one line
+        /// with the numerator directly above the
+        /// demoninator, with the slash ("/") lying horizontally between them.
```
ZevEisenberg: #nth this may be misleading, because at this point it's no longer a slash, but a fully horizontal line. This is just a guess, because I can't find any fonts that support this feature in practice.

Sources/FontFeatures.swift
```diff
@@ -97,6 +97,34 @@
...
+        case vertical
+
+        public func featureSettings() -> [(type: Int, selector: Int)] {
+            switch self {
+            case .none:
```
ZevEisenberg: #req rename to `.disabled`.

Sources/FontInspector.swift
```diff
@@ -170,6 +170,14 @@
 
     }
 
+    extension Fractions: MappableFeature {
```
ZevEisenberg: #req move after `NumberSpacing` for consistency. I think the code may be inconsistent about whether number case or number spacing is first everywhere, but I'd at least like them to stick together.

Tests/AdaptiveStyleTests.swift
```diff
@@ -143,6 +143,9 @@ class AdaptiveStyleTests: XCTestCase {
             (.numberCase(.lower), false),
             (.numberSpacing(.monospaced), false),
             (.numberSpacing(.proportional), false),
+            (.fractions(.disabled), true),
+            (.fractions(.diagonal), false),
+            (.fractions(.vertical), false),
```
ZevEisenberg: #req this causes the test to fail. This array should contain only adjustments of features that EBGaramond12-Regular supports, so delete the `.vertical` line.

Tests/Helpers.swift
```diff
@@ -155,6 +155,7 @@ let fullStyle: StringStyle = {
         fullStyle.fontFeatureProviders = [NumberCase.upper, NumberSpacing.proportional, VerticalPosition.superscript]
         fullStyle.numberCase = .upper
         fullStyle.numberSpacing = .proportional
+        fullStyle.fractions = .disabled
```
ZevEisenberg: #req should be `.diagonal` - non-defaults only in this section.

Example-iOS/DemoStrings.swift
```diff
@@ -306,6 +306,26 @@ enum DemoStrings {
...
+    static let fractionsExample: NSAttributedString = {
+
+        let systemFontStyle = StringStyle(
+            .font(.systemFont(ofSize: 24)),
+            .adapt(.body)
+        )
+
+        let string = "\n1336 <fraction>6/10</fraction> + <fraction>4/10</fraction> = 1337"
+        return NSAttributedString.composed(of: [
+            "Fractions".styled(with: systemFontStyle, .smallCaps(.fromLowercase), .color(.raizlabsRed)),
```
EthanSchatzline: TIL "solidus"

Sources/FontFeatures.swift
EthanSchatzline: how about `with a line lying horizontally between them` or what word would you prefer?

Sources/FontInspector.swift
```diff
@@ -177,6 +177,14 @@
         }
 
     }
+    
+    extension Fractions: MappableFeature {
+
```
ZevEisenberg: #req Whitespace-only lines should have their trailing whitespace trimmed. This one is Xcode's fault - it fails to correctly strip the blank lines when pasting a block like this 😞

EthanSchatzline: Done!

Example-iOS/DemoStrings.swift
```diff
@@ -306,6 +306,24 @@ enum DemoStrings {
...
+    // Demonstrate fractions.
+    static let fractionsExample: NSAttributedString = {
+
+        let garamondStyle = StringStyle(
+            .font(UIFont(name: "EBGaramond12-Regular", size: 24)!),
+            .lineHeightMultiple(1.2),
+            .adapt(.body)
+        )
+
+        let string = "\n1336 <fraction>6/10</fraction> + <fraction>4/10</fraction> = 1337"
```
ZevEisenberg: #req delete `\n`
#pp I think this looks better with uppercase numbers, but uppercase numbers and diagonal fractions don't mix with Garamond, so you'd have to switch back to lower in `<fraction>`:

```swift
let string = "1336 <fraction>6/10</fraction> + <fraction>4/10</fraction> = 1337"
return string.styled(with: garamondStyle.byAdding(
    .numberCase(.upper),
    .xmlRules([
        .style("fraction", garamondStyle.byAdding(.fractions(.diagonal), .color(.raizlabsRed), .numberCase(.lower))),
        ])
    )
)
```

ZevEisenberg:
@EthanSchatzline thanks for your contribution! This will be in a release soon.

PR #275, opened 2017-03-25
String transforms

### LINKED ISSUE #26: Support changing text case
A `BONText` should be able to specify that its string will be transformed to uppercase, lowercase, and title case.

ZevEisenberg:
Fixes #26.

### DIFF
--- a/Sources/StringStyle.swift
+++ b/Sources/StringStyle.swift
@@ -176,19 +176,19 @@ extension StringStyle {
         }
         let tagsApplied = NSAttributedString(string: theString, attributes: supplyDefaults(for: existingAttributes))
 
-        if let transform = transform {
-            let mutable = tagsApplied.mutableStringCopy()
-            let fullRange = NSRange(location: 0, length: mutable.length)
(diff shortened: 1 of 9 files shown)


Sources/StringStyle.swift
```diff
@@ -173,7 +174,21 @@ extension StringStyle {
                 return attributedString
```
ZevEisenberg: #? @KingOfBrian this is working, but how? If we return here, we won't ever get to the transform. I built this the other night, and everything seems to be working, but I can't see how.

KingOfBrian: The XML Builder will call this again and again as it builds the XML string.

Sources/XMLBuilder.swift
```diff
@@ -322,33 +328,54 @@ class XMLBuilder: NSObject, XMLParserDelegate {
         xmlStylers.removeLast()
     }
 
+    func foundNew(string theString: String) {
+        let newAttributedString = topStyle.attributedString(from: theString)
+        attributedString.append(newAttributedString)
+    }
```
KingOfBrian: Should be able to pull more logic in here, a bunch of duplication below.

PR #288, opened 2017-05-05
Update README.md

kviksilver:
Currently README uses badge pointing to default CircleCI png for project, ant that one is showing status of last build. That means when there is a failure in last PR job but master job still passes, badge will be wrong. This changes it to use master branch badge.


ZevEisenberg:
Wow, great workaround! I had given up on having the badge always green. Thanks!

PR #290, opened 2017-05-17
Remove NSAttributedStringKey.kern from last character of styled strings

### LINKED ISSUE #284: Remove NSKernAttributeName from last character
If centered text has tracking applied (nonzero and non-nil `NSKernAttributeName`), the extra spacing is applied to the last character as well, which can make the text look not-centered. To fix this, when text is centered (or always?), we should remove this attribute from the last character. Thanks to @ndonald2 for teaching me about this.

This was supported in BonMot 3.x, so it's a regression that it doesn't work in 4.x.

ZevEisenberg:
fixes #284

- [x] Fix failing test
- [x] wait for #296 to land

### DIFF
--- a/Sources/Composable.swift
+++ b/Sources/Composable.swift
@@ -65,26 +65,30 @@ public extension Composable {
 
     /// Create a new `NSAttributedString` with the style specified.
     ///
-    /// - parameter style:         The style to use.
+    /// - parameter style: The style to use.
     /// - parameter overrideParts: The style parts to override on the base style.
+    /// - parameter stripTrailingKerning: whether to strip NSAttributedStringKey.kern
+    ///                                   from the last character of the result.
     /// - returns: A new `NSAttributedString`.
-    public func styled(with style: StringStyle, _ overrideParts: StringStyle.Part...) -> NSAttributedString {
+    public func styled(with style: StringStyle, _ overrideParts: StringStyle.Part..., stripTrailingKerning: Bool = true) -> NSAttributedString {
         let string = NSMutableAttributedString()
         let newStyle = style.byAdding(stringStyle: StringStyle(overrideParts))
-        append(to: string, baseStyle: newStyle, isLastElement: !Thread.current.isCurrentlyComposing)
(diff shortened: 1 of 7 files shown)


ZevEisenberg:
I got a great suggestion from @fcanas, @danielpunkass, and @mattmassicotte about avoiding thread-local storage: when constructing strings using `styled(with:overrideParts:)`, in cases where we strip the final-character kerning attribute, add our own custom `com.raizlabs.bonmot.kerningAttributeRemovedHere` attribute to the last character of the string. Then, in `composed(of:baseStyle:separator:)`, look for that attribute, and if it's there and it's not the last item, reinstate the kerning attribute. I like this better than thread-local, so I'm going to take a crack at switching to this method.

Oh, and @mattmassicotte pointed out that what I probably wanted in the first place was queue-local storage, not thread-local storage. Good distinction to make!

ZevEisenberg:
The new method also has the nice bonus of working with any string created with `styled(with:overrideParts:)`, instead of just those created with that method _but only when passed to `composed(of:baseStyle:separator)` directly._ I see this as a win.

Sources/Composable.swift
```diff
@@ -208,8 +248,48 @@ extension Special: Composable {
     ///
     /// - parameter to:        The attributed string to which to append the receiver.
     /// - parameter baseStyle: The style to use.
-    public func append(to attributedString: NSMutableAttributedString, baseStyle: StringStyle) {
+    public func append(to attributedString: NSMutableAttributedString, baseStyle: StringStyle, isLastElement: Bool) {
         description.append(to: attributedString, baseStyle: baseStyle)
```
joe-goullaud: Should `isLastElement` be passed along here?

ZevEisenberg: I looked at this. The method is part of the `Composable` protocol, but the `Special` type doesn't care about it, so it ignores it.

mattmassicotte:
👋 I’ll have discussions about TLS any time!

Great work fixing this!

PR #295, opened 2017-08-05
Resume testing Carthage support, force bottle installation.

jdhealy:
Force uninstallation of all `carthage` formulae from `homebrew`.

Avoid second (unnecessary) update, and force installation of `carthage` [bottle (binary package)](https://github.com/Homebrew/brew/blob/master/docs/Bottles.md) from `homebrew`.


ZevEisenberg:
Well I'll be. That's a clever fix.

PR #296, opened 2017-08-06
Add Xcode 9, iOS 11, tvOS 11, watchOS 4, and macOS 10.13 support

ZevEisenberg:


Sources/UIKit/StyleableUIElement.swift
```diff
@@ -224,3 +224,31 @@ internal class StringStyleHolder: NSObject {
...
+        }
+
+        return newDict
+    }
+
+}
+
+extension Dictionary where Key == String {
+
+    func withTypedKeys<KeyType>() -> [KeyType: Value] where KeyType: RawRepresentable, KeyType.RawValue == String {
```
Imperiopolis: These are some super handy helpers :)

Sources/UIKit/StyleableUIElement.swift
```diff
@@ -224,3 +224,31 @@ internal class StringStyleHolder: NSObject {
...
+        var newDict: [KeyType: Value] = [:]
+        for (key, value) in self {
+            if let newKey = KeyType(rawValue: key) {
```
Imperiopolis: Not sure about RZ coding style, but generally I'd do something like this as 

```swift
for (key, value) in self {
    guard let newKey = KeyType(rawValue: key) else { continue }
    newDict[newKey] = value
}
```

Sources/UIKit/UIKit+AdaptableTextContainerSupport.swift
```diff
@@ -37,8 +32,8 @@ extension UIApplication {
     @objc internal func bon_notifyContainedAdaptiveContentSizeContainers(fromNotification notification: NSNotification) {
         // First notify the app delegate if it conforms to AdaptableTextContainer.
         if let container = delegate, let traitCollection = container.window??.traitCollection {
-            if container.responds(to: #selector(AdaptableTextContainer.adaptText(forTraitCollection:))) {
-                container.perform(#selector(AdaptableTextContainer.adaptText(forTraitCollection:)), with: traitCollection)
+            if NSObject.responds(to: #selector(AdaptableTextContainer.adaptText(forTraitCollection:))) {
+                NSObject.perform(#selector(AdaptableTextContainer.adaptText(forTraitCollection:)), with: traitCollection)
```
Imperiopolis: Curious why the change to `NSObject` here? Also, would probably reduce the if to `if let traitCollection = delegate?.window??.traitCollection {` since you're no longer using container

Imperiopolis:
Also, best kind of PRs 🎉 
image

Sources/UIKit/StyleableUIElement.swift
ZevEisenberg: I don't remember writing these! Looks like I did, though, unless I found them somewhere and forgot to give credit. But I'm usually pretty good about that when pasting code.

Sources/UIKit/StyleableUIElement.swift
ZevEisenberg: I don't think our style addresses it specifically. (We don't really have an official style guide like we did with Obj-C. We just defer to the .swiftlint.yml from our most recent project, and we try to optimize client code for readability and eventual client handoff ¯\\\_(ツ)_/¯

Sources/UIKit/UIKit+AdaptableTextContainerSupport.swift
ZevEisenberg: I don't know why this was changed. Guessing the migrator made a mistake. I can't think of any reason to keep it, and it looks like it would break things, so I'm going to change it back.

ZevEisenberg:
@Imperiopolis thanks so much for the review! 🚀

Issue #297, opened 2017-08-09
StringStyle attributes should contain the transform


jvisenti:
I have a case where I'm creating a `StringStyle`, retrieving it's `attributes`, and caching those for later manual construction of an `NSAttributedString`. The `StringStyle` in question has a custom `transform` set.

Expected:
`attributes` dictionary contains a value for `BonMotTransformationsAttributeName` key, which can be used to apply the desired transform to strings manually

Actual:
There is no representation of the `StringStyle`'s transform in its `attributes` dictionary

jvisenti:
Also, it looks like `transformer` is `internal` on `Transform`, so they can't be applied manually.

PR #298, opened 2017-08-09
Support manual application of StringStyle transforms via attributes dictionary

jvisenti:
Addresses #297

### DIFF
--- a/Sources/Transform.swift
+++ b/Sources/Transform.swift
@@ -25,7 +25,7 @@ public enum Transform {
     case capitalizedWithLocale(Locale)
     case custom(TransformFunction)
 
-    var transformer: TransformFunction {
+    public var transformer: TransformFunction {
         switch self {
         case .lowercase: return { string in string.localizedLowercase }
         case .uppercase: return { string in string.localizedUppercase }
--- a/Sources/Transform.swift
+++ b/Sources/Transform.swift
@@ -41,73 +41,75 @@ public enum Transform {
 
 }
 
-extension Transform: EmbeddedTransformation {
+#if os(iOS) || os(tvOS)
+    extension Transform: EmbeddedTransformation {
 
-    internal enum Keys {
+        internal enum Keys {
 
-        static let name = "name"
-        static let locale = "locale"
+            static let name = "name"
+            static let locale = "locale"
 
-    }
-
-    internal enum Name: String {
-
-        case lowercase
-        case uppercase
-        case capitalized
-        case custom
-
-    }
-
-    internal var name: Name {
-        switch self {
-        case .lowercase, .lowercaseWithLocale: return Name.lowercase
-        case .uppercase, .uppercaseWithLocale: return Name.uppercase
-        case .capitalized, .capitalizedWithLocale: return Name.capitalized
-        case .custom: return Name.custom
-        }
-    }
-
-    public var asDictionary: StyleAttributes {
-        var dict: StyleAttributes = [
-            Keys.name: name.rawValue,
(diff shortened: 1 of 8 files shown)


ZevEisenberg:
@jvisenti have you been using this in production code? Think it's up to snuff for merging?

ZevEisenberg:
Sounds reasonable. We should rebase this at some point, or port the changes over, now that #296 has landed. Rebase looks a little hairy, so a by-hand port might make more sense.

Sources/Transform.swift
```diff
@@ -25,7 +25,7 @@ public enum Transform {
     case capitalizedWithLocale(Locale)
     case custom(TransformFunction)
 
-    var transformer: TransformFunction {
+    public var transformer: TransformFunction {
```
cjamie: this change shouldn't be necessary

Sources/Transform.swift
```diff
@@ -40,3 +40,76 @@ public enum Transform {
...
+            case lowercase
+            case uppercase
+            case capitalized
+            case custom
+
+        }
+
+        internal var name: Name {
+            switch self {
+            case .lowercase, .lowercaseWithLocale: return Name.lowercase
```
cjamie: would like to take advantage of swift's type inference: 
```suggestion 
            case .lowercase, .lowercaseWithLocale: return .lowercase
```

Tests/AttributedStringStyleTests.swift
```diff
@@ -37,7 +37,7 @@ class StringStyleTests: XCTestCase {
     #if os(iOS) || os(tvOS)
     func testTextStyle() {
         let style = StringStyle(.textStyle(titleTextStyle))
-        for (style, fullStyle) in additiviePermutations(for: style) {
+        for (style, fullStyle) in additivePermutations(for: style) {
```
cjamie: while we're here, why don't we prefer forEach instead? 
```suggestion 
       additivePermutations(for: style).forEach { (style, fullStyle) in
```

Sources/UIKit/EmbeddedTransformation.swift
```diff
@@ -60,7 +60,7 @@ internal enum EmbeddedTransformationHelpers {
         let representations = styleAttributes[BonMotTransformationsAttributeName] as? [StyleAttributes] ?? []
         let results: [T?] = representations.map { representation in
             for type in embeddedTransformationTypes {
-                if let transformation = type.from(dictionary: representation) as? T {
+                if let transformation = type.init(dictionary: representation) as? T {
```
cjamie: for line 61, we can change this to a compactMap so we don't need to go through the collection again, and with flattop
```suggestion
         let results: [T] = representations.compactMap { representation in
```

Tests/TransformTests.swift
```diff
@@ -183,4 +183,25 @@ class TransformTests: XCTestCase {
...
+            .adapt(.control)
+        )
+
+        let attributes = style.attributes
+        XCTAssertNotNil(attributes[BonMotTransformationsAttributeName])
+
+        let transform = (attributes[BonMotTransformationsAttributeName] as? StyleAttributes).flatMap(Transform.init)
+        XCTAssertNotNil(transform)
+
+        if let transformer = transform?.transformer {
```
cjamie: in tests, I think its ok to force unwrap, so we can just get a failing test

chrisballinger:
Looks like this has been hanging open for a few years. Going to close this for now, but if someone needs this feature and wants to resurrect it, go ahead!

Issue #251, opened 2016-11-24
Support use in app extensions


ZevEisenberg:
See [this similar request for Anchorage](https://github.com/Raizlabs/Anchorage/pull/21#issuecomment-262791384). When I enable the `APPLICATION_EXTENSION_API_ONLY` flag, I get this:

```BonMot/Sources/UIKit/UIKit+Helpers.swift:61:26: error: 'shared' is unavailable: Use view controller based solutions where appropriate instead.
        if UIApplication.shared.delegate != nil {
                         ^~~~~~
UIKit.UIApplication:5:20: note: 'shared' has been explicitly marked unavailable here
    open class var shared: UIApplication { get }
                   ^
BonMot/Sources/UIKit/UIKit+Helpers.swift:62:34: error: 'shared' is unavailable: Use view controller based solutions where appropriate instead.
            return UIApplication.shared.preferredContentSizeCategory
                                 ^~~~~~
UIKit.UIApplication:5:20: note: 'shared' has been explicitly marked unavailable here
    open class var shared: UIApplication { get }
                   ^
```

soundbites:
I had the same issue yesterday. I've sorta fixed with a pre-processor
```
        // `UIApplication.shared` is not a valid object in unit tests. Fall back
        // to a default value if the delegate is nil.
        // Also make UIApplication is not available in extensions
#if !TARGET_IS_EXTENSION
        if UIApplication.shared.delegate != nil {
            return UIApplication.shared.preferredContentSizeCategory
        }
#endif
        
        #if swift(>=3.0)
            return UIContentSizeCategory.large
        #else
            return UIContentSizeCategoryLarge
        #endif
```

and then setting up the `other Swift flags` with `-DTARGET_IS_EXTENSION`

ZevEisenberg:
@soundbites don't extensions support Dynamic Type? Isn't there a global function to get the content size category, or some other way to get it in extensions? It looks like your solution would just return a constant value in extensions, if I'm reading it right.

ZevEisenberg:
I posted a [Stack Overflow question](http://stackoverflow.com/questions/41388934/equivalent-of-uiapplication-shared-preferredcontentsizecategory-in-app-extension) on how to solve this. Looking for ways to get the extension's preferred trait collection.

LeoNatan:
Does `UIScreen.main.preferredContentSizeCategory` return an expected value? Iirc, `UIScreen.main` is officially available in extensions.

[...]

PR #301, opened 2017-09-10
Allow BonMot to be used in extensions

plarson:


ZevEisenberg:
@plarson thanks for this! Does this also resolve https://github.com/Raizlabs/BonMot/issues/251?

Sources/UIKit/UIKit+Helpers.swift
```diff
@@ -56,18 +56,27 @@ extension UITraitCollection {
...
-        }
-        else {
+
+        #if NS_EXTENSION_UNAVAILABLE_IOS
+            // `UIApplication.shared` is not a valid object in unit tests. Fall back
+            // to a default value if the delegate is nil.
+            if UIApplication.shared.delegate != nil {
+                return UIApplication.shared.preferredContentSizeCategory
+            }
+            else {
```
ZevEisenberg: I think you could drop the `else` and the `#else`, because if we make it into this `#if`, we will return, and if not, we always want to return `UIContentSizeCategory.large`.

plarson:
Sorry my bad, that macro doesn't do what it seems like it does

teameh:
This did end up in another PR?

ZevEisenberg:
I don’t think so 😞. @teameh are you running into issues with BonMot in extensions? If so, could you please leave a comment on #251 or open a new issue?

teameh:
Okay, but in theory the fix for this is quite small right? I'll add a new PR

lammertw:
@teameh did you manage to fix it?

teameh:
Nope.. But I did dig a bit. See https://github.com/Rightpoint/BonMot/issues/251

PR #302, opened 2017-09-22
New accessibility attributes

### LINKED ISSUE #292: Add support for new accessibility attributes
Announced at WWDC 2017:

[`UIAccessibilitySpeechAttributeQueueAnnouncement`](https://developer.apple.com/documentation/uikit/uiaccessibilityspeechattributequeueannouncement?changes=latest_minor)
[`UIAccessibilitySpeechAttributeIPANotation `](https://developer.apple.com/documentation/uikit/uiaccessibilityspeechattributeipanotation?changes=latest_minor)

Documentation is thin on the ground as of this writing, so this is a reminder to check later to see if it's been filled in.

ZevEisenberg:
Fixes #292

- [x] wait for #296 to merge

### DIFF
--- a/Tests/AttributedStringStyleTests.swift
+++ b/Tests/AttributedStringStyleTests.swift
@@ -127,6 +127,16 @@ class StringStyleTests: XCTestCase {
         }
     }
 
+    func testShouldQueueSpeechAnnouncement() {
+        if #available(iOS 11, tvOS 11, watchOS 4, *) {
+            let style = StringStyle(.shouldQueueSpeechAnnouncement(true))
+            for (style, fullstyle) in additiviePermutations(for: style) {
+                XCTAssertTrue(fullstyle == true || style.attributes.count == 1)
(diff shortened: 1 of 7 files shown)


Tests/AttributedStringStyleTests.swift
```diff
@@ -117,6 +117,36 @@ class StringStyleTests: XCTestCase {
         }
     }
 
+    func testPronuciationStyle() {
+        if #available(iOS 11, tvOS 11, watchOS 4, *) {
```
Imperiopolis: Wonder if there should be a convenience for this `#available` set since using it all over.

ZevEisenberg: Good point. I forgot I could just wrap this in a function/closure. I'll check it out and see where it makes sense to add it.

ZevEisenberg: Doesn't look like there's a way to do this that increases clarity, and it obfuscates the intent a little bit.

PR #306, opened 2017-10-15
Use multiline strings in readme and example app.

ZevEisenberg:


Example-iOS/DemoStrings.swift
```diff
@@ -30,7 +30,10 @@ enum DemoStrings {
     static let xmlExample: NSAttributedString = {
 
         // This would typically come from NSLocalizedString
-        let localizedString = "I want to be different. If everyone is wearing <black><BON:noBreakSpace/>black,<BON:noBreakSpace/></black> I want to be wearing <red><BON:noBreakSpace/>red.<BON:noBreakSpace/></red>\n<signed><BON:emDash/>Maria Sharapova</signed> <racket/>"
+        let localizedString = """
+            I want to be different. If everyone is wearing <black><BON:noBreakSpace/>black,<BON:noBreakSpace/></black> I want to be wearing <red><BON:noBreakSpace/>red.<BON:noBreakSpace/></red>
```
ZevEisenberg: I could have done some soft wrapping here, but with all the slashes already, I thought it made things more confusing, not less.

ZevEisenberg:
@timvermeulen suggested this, and I wonder if GitHub will let me assign it after doing a username mention?

Example-iOS/DemoStrings.swift
```diff
@@ -386,26 +394,22 @@ enum DemoStrings {
...
+                """
+                <el-GR><emoji>🇬🇷</emoji> Εδώ είναι ο σωστός τρόπος για να προφέρει Ελληνικά.</el-GR>
+                <en-US><emoji>🇺🇸</emoji> Here is the correct way to pronounce American English.</en-US>
+                <es-DO><emoji>🇩🇴</emoji> Aquí está la manera correcta de pronunciar el español.</es-DO>
+                <fr-FR><emoji>🇫🇷</emoji> Voici comment prononcer correctement le français.</fr-FR>
+                <pt-PT><emoji>🇵🇹</emoji> Aqui está a maneira correta de pronunciar português.</pt-PT>
+                <ru-RU><emoji>🇷🇺</emoji> Вот как правильно произносить русский.</ru-RU>
+                <uk-UA><emoji>🇺🇦</emoji> Ось як правильно вимовляти Український.</uk-UA>
+                <zh-TW><emoji>🇹🇼</emoji> 这里是正确的做法发音中文.</zh-TW>
+                """,
```
timvermeulen: This adds newlines where you previously had none, so depending on whether you want them or not, you can end lines with a `\` to remove the newline 🙂

ZevEisenberg: Good catch. It turns out, the VoiceOver pronunciation hinting system is pretty broken, and if I have the spaces and new lines set up any way other than what was here before, with a newline before `<emoji>` and a space after `</emoji>`, VoiceOver says "Space" in the default Siri voice after reading each sentence. Since multiline string literals actually make the correct arrangement _harder_ to read, I'm going to revert this section.

timvermeulen: 👍

PR #310, opened 2017-11-08
Add Swift version in the podspec file

Dimillian:


ZevEisenberg:
@Dimillian thanks for your contribution! Just so I can track this, and help future people who may have questions, what problem does this solve?

ZevEisenberg:
Follow up: I understand the PR, but since CocoaPods 1.4.0, you can also add the Swift version explicitly using `swift_version`. Should we just use that? Can we use both?

Dimillian:
@ZevEisenberg I'm very sorry for the late reply. I didn't really had time to dig inside latest Cocoapods changes, but if you can specify a new property, forget about my PR and use the new cocoa pods property. I never was a fan of modifying project properties from the pod spec anyway.

PR #313, opened 2017-11-14
Return current BONFont, if has no features

muukii:
I found overhead in creating a font.
I think the creating font should return `self` if it has no features.
So, StringStyle will be faster.

What do you think?


ZevEisenberg:
@muukii cool, thanks for this contribution! I ran a couple of quick tests, and it does indeed seem to be faster. I wish there were a convenient way to add those performance benchmarks to the tests, so that we don't regress, but for now I'm just happy to have this improvement 🎉

PR #322, opened 2018-02-08
Fixes #320. Fixes compiler errors and warnings encountered with Xcode 9.3 beta 2

### LINKED ISSUE #320: Compile error using XCode 9.3 beta
Error message: "Binary operator '+=(_:_:)' cannot be applied to operands of type ... 
At line# 136 in file StringStyle.swift 

Br,

rdingman:


ZevEisenberg:
@rdingman thanks for this PR! I've got one little tweak that I'd love to see:

I prefer to avoid having Swift-version-specific conditionals in the main code wherever possible. Instead, I like to back-port new functions to the old version of Swift. We wrote up a [blog post](https://www.raizlabs.com/dev/2016/10/swift-2-x-3-0-compatibility/) about this, but the short version is that, in the existing Compatibility.swift, we would expose a function called `compactMap` only on Swift 4.0. This way, we don't have two branches in StringStyle.swift that we have to keep in sync with each other.

How does that sound? Please let me know if you'd like clarification, or if you don't have time to make the update, I can take it from here.

rdingman:
Makes sense. I'll try to make those changes today.

ZevEisenberg:
@rdingman perfect. Going to wait a little to merge until the Xcode is a little further along, because future betas may break more things, but I may publish this as a release candidate soon.

ZevEisenberg:
This has been released as part of [BonMot 5.1](https://github.com/Raizlabs/BonMot/releases/tag/5.1). Thanks for your contribution, @rdingman! 🚀

PR #324, opened 2018-02-13
Transform 'kerning(forFont: )' method to public

acacuce:
Currently kerning(forFont: ) method in Tracking enum has an internal attribute. It would be good if it will be public

### DIFF
--- a/Sources/Tracking.swift
+++ b/Sources/Tracking.swift
@@ -14,7 +14,7 @@ public enum Tracking {
     case point(CGFloat)
     case adobe(CGFloat)
 
-    func kerning(forFont font: BONFont?) -> CGFloat {
+    public func kerning(forFont font: BONFont?) -> CGFloat {
         switch self {
         case .point(let kernValue):
             return kernValue
--- a/Tests/AccessTests.swift
+++ b/Tests/AccessTests.swift
@@ -11,11 +11,20 @@ import XCTest
 
 class AccessTests: XCTestCase {
 
+    override func setUp() {
+        super.setUp()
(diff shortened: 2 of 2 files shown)


ZevEisenberg:
@acacuce thanks for your pull request! Just curious - what's your use case for making this method public?

acacuce:
I use StringStyles to describe the text styles in our application. Our designers use custom Tracking. This text is used in a chat where the sizes of chat bubble are calculated in with the help of NSTextContainer. For the correct calculation, I need to transfer the value of the tracking value.

acacuce:
As an example, I need to count the hashValue from the value of Tacking

ZevEisenberg:
Ah, yep, that makes sense. Would you mind adding a unit test that confirms that this method stays public? It would be a simple test case that does `import BonMot` instead of `@testable import BonMot`, and then it just needs to call `kerning(forFont:)`, so if the method ever loses `public`, the test won't compile.

Also, as a current workaround, you can get the kerning if you query `yourStringStyle.attributes[.kerning] as? NSNumber` (I think `as? CGFloat` would work too).

ZevEisenberg:
I merged this, but then I realized it should really be spelled `kerning(for:)`, so the call site looks like `kerning(for: font)`. I'm going to push a change for this. Just a heads up in case you're depending on this method name.

ZevEisenberg:
This has been released as part of [BonMot 5.1](https://github.com/Raizlabs/BonMot/releases/tag/5.1). Thanks for your contribution, @acacuce! 🚀

PR #325, opened 2018-02-13
Nested emphases

### LINKED ISSUE #303: Support for mixing bold/italic fonts with tags
Are there any plans to support nested tags? This would be very useful to apply bold+italic for example:

"I'd like to call <b><i>your</i> attention to this</b>."

Thanks,

ZevEisenberg:
Fixes #303.

Would love feedback on naming and API in general.

~add example to readme probably?~ decided not to; it's a nice detail that works when you want it, but no need to call it out explicitly, I think.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -3,6 +3,7 @@ disabled_rules: # rule identifiers to exclude from running
   - nesting
   - identifier_name
   - large_tuple
+  - file_length
 
 statement_position:
   statement_mode: uncuddled_else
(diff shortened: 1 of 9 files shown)


heyltsjay:
I think `Emphasis` is pretty good! Falls apart for things like `.tightLineSpacing` though. 🤔

ZevEisenberg:
I could separate that into a distinct `FontLineSpacingVariant` enum. Makes the API cleaner, but harder to find things if you're coming from knowledge of `NSAttributedString`.

ZevEisenberg:
Gong to try simplifying this to cover just bold and italics for now, since they are by far the most common uses of this feature. New items can be added later with a different API.

ZevEisenberg:
Also, that makes the naming much easier: if it's just `.bold` and `.italic`, then `Emphasis` makes more sense.

PR #329, opened 2018-03-01
Join sequence of composable into attribute string

### LINKED ISSUE #328: Join array of attributed strings with a Composable
It would be nice to have this:

```swift
public extension Sequence where Element: Composable {

    public func joined(separator: Composable) {
        ...
    }

}
```

ZevEisenberg:
Closes #328

### DIFF
--- a/Tests/ComposableTests.swift
+++ b/Tests/ComposableTests.swift
@@ -204,6 +204,10 @@ class ComposableTests: XCTestCase {
             ])
 
         XCTAssertEqual(testWithDefaultSeparator, controlWithDefaultSeparator)
+
+        XCTAssertEqual(["a", "b", "c"].joined(), "abc")
+
+        XCTAssertEqual(["a", "b", "c"].joined() as NSAttributedString, "abc".styled(with: StringStyle()))
     }
 
 }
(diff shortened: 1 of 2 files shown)


ZevEisenberg:
@jvisenti I pushed an additional commit. Could you please have a look at it? I'm wondering whether it's confusing to have `["a", "b"].joined()` be able to produce either a `String` or an `NSAttributedString` depending on type inference. cc @KingOfBrian

Tests/ComposableTests.swift
```diff
@@ -173,4 +173,41 @@ class ComposableTests: XCTestCase {
         XCTAssertEqual(paragraphEnd.lineSpacing, 0)
     }
 
+    func testJoiningSequence() {
+        let redStyle = StringStyle(.color(.red))
+        let greenStyle = StringStyle(.color(.red))
+        let blueStyle = StringStyle(.color(.red))
```
KingOfBrian: Strange shade of blue / green.

ZevEisenberg: d'oh, thank you

KingOfBrian:
Good call on the oddity of having the overloads available, but I think it's fine. It may have an impact on compile times, but I doubt it's significant. It could be surprising to a user, but I'm not sure it's a big gotcha. I wonder if there's any guidance on providing overloaded methods.

jvisenti:
I think in general the context will define the type and there won't be much ambiguity

ZevEisenberg:
I also confirmed that, without any other context, `["a", "b"].joined()` returns a `String`, so we should be OK.

PR #340, opened 2018-06-01
Readme updates

### LINKED ISSUE #339: Scaling Behavior graph does not plot the Apple Dynamic Type behaviors as implied
The scaling behaviors graph is missing `UIFontTextStyleLargeTitle` and `UIFontTextStyleTitle1` properties. Could these be added to the graph? The graph is very helpful.

image

ZevEisenberg:
Fixes #339, plus one other minor fix for something that's been bothering me. @thisisjeffwong I'd appreciate a review of this, if you have time.


ZevEisenberg:
Code I used to generated this, in case its useful to someone, including me from the future: 
[Dynamic Type Research.zip](https://github.com/Raizlabs/BonMot/files/2064101/Dynamic.Type.Research.zip)

thisisjeffwong:
The graph shows most of the styles not growing in the larger accessibility sizes. However, I noticed that they still do when looking at the [Apple typography docs](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography#dynamic-type-sizes). 

I'm not sure which is correct though it doesn't make sense to me why they would have some sizes simply not grow. 

I didn't notice the discrepancy before when I filed the ticket.

ZevEisenberg:
I only updated the two new sizes. I guess I should update all the sizes, just in case they’ve changed. I really should throw together a script that can do this all via a graphing library or something.

cjamie:
can we also include changing the `struct Constants` in this PR to `enum Constants` since its more appropriate as a namespace

ZevEisenberg:
@cjamie which constants are those? I don’t think it’s related to this PR; would you mind opening a new PR or issue with more detail?

cjamie:
@ZevEisenberg  the ones in the readme

PR #360, opened 2019-04-11
Consider emoji when removing kerning from last character

LorDisturbia:
The current implementation of `removeKerningFromLastCharacter` and `restoreKerningOnLastCharacter` start from the assumption that the last character range always has length 1. This however is not true for more complex characters, such as emojis.
The consequence of this was a bug in which applying a style to a string that ended with an emoji, while keeping the flag `stripTrailingKerning` to its default value of `true`, would result in the emoji not being shown at all.

### DIFF
--- a/Sources/Composable.swift
+++ b/Sources/Composable.swift
@@ -275,7 +275,8 @@ extension NSMutableAttributedString {
             return
         }
 
-        let lastCharacterRange = NSRange(location: length - 1, length: 1)
+        let lastCharacterLength = string.suffix(1).utf16.count
+        let lastCharacterRange = NSRange(location: length - lastCharacterLength, length: lastCharacterLength)
 
         guard let currentKernValue = attribute(.kern, at: lastCharacterRange.location, effectiveRange: nil) else {
             return
@@ -290,7 +291,8 @@ extension NSMutableAttributedString {
             return
         }
 
-        let lastCharacterRange = NSRange(location: length - 1, length: 1)
+        let lastCharacterLength = string.suffix(1).utf16.count
+        let lastCharacterRange = NSRange(location: length - lastCharacterLength, length: lastCharacterLength)
 
         guard let currentKernValue = attribute(.bonMotRemovedKernAttribute, at: lastCharacterRange.location, effectiveRange: nil) else {
             return
--- a/Tests/ComposableTests.swift
+++ b/Tests/ComposableTests.swift
@@ -213,23 +213,22 @@ class ComposableTests: XCTestCase {
     func testKerningStrippingOnLastCharacter() {
         let styleWithTracking = StringStyle(.tracking(.point(0.5)))
 
-        let testCases: [String] = [
-            "abc",
-            "abc🍕",
-            "🍕🍕",
+        let pairsline = #line; let testCases: [(input: String, lastCharacterUnicodeLength: Int)] = [
+            ("abc", 1),
+            ("abc🍕", 2),
+            ("🍕🍕", 2),
         ]
 
-        for testCase in testCases {
-            let styledString = testCase.styled(with: styleWithTracking)
+        for (offset, testCase) in testCases.enumerated() {
+            let styledString = testCase.input.styled(with: styleWithTracking)
 
             var range = NSRange(location: 0, length: 0)
             let maxRange = NSRange(location: 0, length: styledString.length)
 
             let kerning = styledString.attribute(.kern, at: 0, longestEffectiveRange: &range, in: maxRange) as? Float
-
-            let lastCharacterUnicodeLength = testCase.suffix(1).utf16.count
-            XCTAssertEqual(kerning, 0.5)
-            XCTAssertEqual(range, NSRange(location: 0, length: styledString.length - lastCharacterUnicodeLength))
+            let lineNumber = pairsline + offset + 1
+            XCTAssertEqual(kerning, 0.5, line: UInt(lineNumber))
+            XCTAssertEqual(range, NSRange(location: 0, length: styledString.length - testCase.lastCharacterUnicodeLength), line: UInt(lineNumber))
         }
     }
 }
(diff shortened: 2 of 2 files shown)


Sources/Composable.swift
```diff
@@ -275,7 +275,8 @@ extension NSMutableAttributedString {
             return
         }
 
-        let lastCharacterRange = NSRange(location: length - 1, length: 1)
+        let lastCharacterLength = string.suffix(1).utf16.count
```
ZevEisenberg: Pretty sure `suffix` is a O(n) operation. I'm wondering if there's something we can do something like `.reversed.first` that will count from the end and still get the extended grapheme clusters right?

ZevEisenberg:
@LorDisturbia thank you so much for reporting this! I'd love to have it get fixed and _stay_ fixed. Would you mind adding some tests?

Sources/Composable.swift
LorDisturbia: @ZevEisenberg I just run some tests and it looks that `.suffix`, `.reversed.first` and `last` are all O(1) operations. I couldn't find any official documentation about it, though.

On the other hand, now that I though about it, I would probably go for something like
```guard length != 0, let lastCharacter = string.last else {
            return
        }

        let lastCharacterLength = String(lastCharacter).utf16.count
        let lastCharacterRange = NSRange(location: length - lastCharacterLength, length: lastCharacterLength)```

just because it looks less hackish than using suffix. But maybe I'm overthinking it.

Comment (LorDisturbia) [reply Sources/Composable.swift]: @ZevEisenberg I just run some tests and it looks that .suffix, .reversed.first and last are all O(1) operations. I couldn't find any official documentation about it, though.

The syntax advantage of suffix is that it is automatically coalesced with an empty string, while for first and last we would have to unwrap it (even though we already know the string is longer than 0 characters)

**Comment by LorDisturbia** [comment]
Hi @ZevEisenberg, I updated the PR with a test. :)

**Comment by LorDisturbia** [comment]
Let me know if need anything else from me in order to merge the PR. :)

**Comment by LorDisturbia** [comment]
@ZevEisenberg Let me know if need anything else from me in order to merge the PR. :)

**Comment by ZevEisenberg** [comment]
@LorDisturbia will do - I'll need a little time to review it, which I haven't had yet. Is this blocking you? Can you point your project at the branch? I'll try to get it landed when I've got some time. Thanks for adding tests!

**Comment by LorDisturbia** [comment]
@ZevEisenberg No worries, just wanted to check if you needed anything else. We already implemented a workaround in our project. :)

**Comment by ZevEisenberg** [comment]
Thanks! We've got a new co-op who is clearing some of the dust off BonMot and getting it building properly in the latest Xcode. Once we do that, we'll hopefully have some time to merge in some of the PRs that have been piling up. I'm excited to land this one!

File: Sources/Composable.swift (hunk 1 above)
Comment (ZevEisenberg) [reply Sources/Composable.swift]: You are correct. Not sure what I was thinking. `suffix` iterates backwards through the UTF-8 or UTF-16 code points (depending on Swift version) until it finds the start of a `Character`, so it will be O(number of unicode scalars of last `Character`). Or something. I’m probably misusing the term for the things it’s looping through, but that’s the gist.

File: Tests/ComposableTests.swift [hunk 2]
```diff
@@ -210,4 +210,26 @@ class ComposableTests: XCTestCase {
         XCTAssertEqual(["a", "b", "c"].joined() as NSAttributedString, "abc".styled(with: StringStyle()))
     }
 
+    func testKerningStrippingOnLastCharacter() {
+        let styleWithTracking = StringStyle(.tracking(.point(0.5)))
+
+        let testCases: [String] = [
```
ZevEisenberg: Let’s capture the line number of this, calculate the index of each test in the array, and pass it to `XCTAssertEqual` methods so that failures occur on the correct lines. This technique is used in a couple places in BonMot’s tests, and is outlined here: https://www.rightpoint.com/rlabs/2019/03/xctest-line-by-line/

Tests/ComposableTests.swift
```diff
@@ -210,4 +210,26 @@ class ComposableTests: XCTestCase {
...
+        for testCase in testCases {
+            let styledString = testCase.styled(with: styleWithTracking)
+
+            var range = NSRange(location: 0, length: 0)
+            let maxRange = NSRange(location: 0, length: styledString.length)
+
+            let kerning = styledString.attribute(.kern, at: 0, longestEffectiveRange: &range, in: maxRange) as? Float
+
+            let lastCharacterUnicodeLength = testCase.suffix(1).utf16.count
+            XCTAssertEqual(kerning, 0.5)
```
ZevEisenberg: When testing floating-point equality, we should be using the `accuracy:` parameter to avoid floating point precision issues. You can see it in a few places in the repo. There’s no hard and fast rule for how small to make the accuracy value, but for kerning, I’d say 0.001 is probably sufficient. Maybe 0.0001 just to be safe.

LorDisturbia:
I don't mind at all! I'm happy to know my fix is going to be included :)

Tests/ComposableTests.swift
```diff
@@ -210,4 +210,25 @@ class ComposableTests: XCTestCase {
...
+        ]
+
+        for (offset, testCase) in testCases.enumerated() {
+            let styledString = testCase.input.styled(with: styleWithTracking)
+
+            var range = NSRange(location: 0, length: 0)
+            let maxRange = NSRange(location: 0, length: styledString.length)
+
+            let kerning = styledString.attribute(.kern, at: 0, longestEffectiveRange: &range, in: maxRange) as? Float
+            let lineNumber = pairsline + offset + 1
```
ZevEisenberg: Style: I’d prefer making this the `Uint`, so the lines where you pass them are shorter. Also, I usually call this `pairLine` or `testLine`.

PR #363, opened 2019-05-15
Fix issue where test result bundle could not be found

chrisballinger:
Not the cleanest solution, but I wanted to run this on CI and see if this fix works there as well.


ZevEisenberg:
@chrisballinger were all the project and scheme changes intentional? We're now three Inception-dream-levels deep into this PR. It goes

master
→ swift-5
  → fastlane-xcov-fix
    → xcode_10.2_fixes

I'd love to collapse this a little more cleanly, just to keep track of where the various changes belong. If it's too much of a hassle to disentangle things, that's fine, but if it's relatively easy, I'd like to keep the commits small and atomic.

fastlane/Fastfile
```diff
@@ -140,6 +137,29 @@ def coverage(scheme:, devices: nil, destination: nil)
     cobertura_xml: "true",
     build_directory: DERIVED_DATA_PATH
   )
+
+  # Since xchtmlreport depends on the result bundle path, we need to run the tests again
+  # AFTER xcov has run, because running it before causes issues finding the test logs in derived data
```
ZevEisenberg: #? Are there any bugs we can link to or anything like that? Obviously, I'd love not to have to run the tests twice, and if we can't avoid that today, I'd like to at least keep an eye toward the future at least.

chrisballinger: I couldn't find any documentation about a change in how Xcode treats the `-resultBundlePath` flag, but I think that's the root of the issue. My hypothesis is that in previous versions of Xcode the test result bundle was available in both derived data as well as the custom path, which is why it is no longer present.

I am going to try one more thing later today to see if we can eliminate the duplicate test run, and might report this Xcode behavior change upstream to the fastlane scan docs.

chrisballinger:
Haha yeah we are a few PR levels deep, but I didn't want to just start adding test commits to the other PRs. I'd also like to see if I can eliminate the duplicate test run before we merge.

PR #364, opened 2019-05-22
Cleanup CircleCI

chrisballinger:
* Uses CocoaPods 1.7.0rc2 for Swift 4.2 + 5.0 podspec support
* CocoaPods 1.7 Specs CDN
* Use CircleCI 2.1 executors and commands

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,39 +1,10 @@
-version: 2
+version: 2.1
 
 jobs:
   danger:
-    macos:
-      xcode: "10.2.1"
-    environment:
-      LC_ALL: en_US.UTF-8
-      LANG: en_US.UTF-8
-      BUNDLE_JOBS: 3
-      BUNDLE_RETRY: 3
-      BUNDLE_PATH: /Users/distiller/project/build/gems
-      HOMEBREW_NO_AUTO_UPDATE: 1
-    shell: /bin/bash --login -eo pipefail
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -182,43 +87,56 @@ jobs:
...
+            - vendor/bundle
+
+  lint-pod:
+    description: "Lints podspec with specified Swift version"
+    parameters:
+      swift-version:
+        type: string
+        default: "5.0"
+    steps:
+      - run: bundle exec pod repo add-cdn jsDelivr-Specs "https://cdn.jsdelivr.net/cocoa/"
```
ZevEisenberg: #? why this repo, in particular?

chrisballinger: It's the only Specs CDN so far: http://blog.cocoapods.org/CocoaPods-1.7.0-beta/

ZevEisenberg: Ohh, got it. I thought it was going to be a system where any company could sign up to host a mirror, but it’s actually just one free CDN, which is sponsored by a bunch of companies. Smart. https://www.jsdelivr.com

chrisballinger: My guess is that once it's no longer experimental there might be some other CDN mirrors pop up. CDN Specs still doesn't support some use cases like `trunk push`, as far as I'm aware.

PR #365, opened 2019-05-22
Add static framework targets for Carthage users

chrisballinger:
These targets are duplicates of their dynamic counterparts, with the only difference being `MACH_O_TYPE = staticlib`.

This is an "officially supported" configuration for Carthage 0.30 and higher. More info here: https://github.com/Carthage/Carthage#build-static-frameworks-to-speed-up-your-apps-launch-times


ZevEisenberg:
Is there anything we can do with xcconfig files to reduce duplication and maintenance cost here?

chrisballinger:
@ZevEisenberg That's a good idea! I'll give it a shot and see if Carthage picks up on it.

p.s. can we turn off the "This branch is out-of-date with the base branch" feature? It can make it difficult to merge forked PRs if we don't have write access to their remote.

ZevEisenberg:
Done!

PR #373, opened 2019-07-30
Update dependencies

ZevEisenberg:
Fixes a security vulnerability flag we were getting. Also cleans up the Gemfile because some things we were waiting for have been released.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -135,7 +135,7 @@ commands:
         type: string
         default: "5.0"
     steps:
-      - run: bundle exec pod repo add-cdn jsDelivr-Specs "https://cdn.jsdelivr.net/cocoa/"
+      - run: bundle exec pod repo add-cdn CocoaPods-Specs "https://cdn.cocoapods.org/"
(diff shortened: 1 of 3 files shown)


ZevEisenberg:
You can ignore the apparent failures here. They’re due to an issue with the Twitter homepage redesign failing the lint check for URL reachability: https://github.com/CocoaPods/CocoaPods/issues/9049

I guess I’ll have to pass `--allow-warnings` so the pod release can happen. There should be a fix in CocoaPods 1.8 that fixes this, but it’s not even in beta yet.

.circleci/config.yml
```diff
@@ -136,7 +136,10 @@ commands:
         default: "5.0"
     steps:
       - run: bundle exec pod repo add-cdn jsDelivr-Specs "https://cdn.jsdelivr.net/cocoa/"
```
chrisballinger: The jsdelivr repo is no longer maintained, we should probably remove that line or change it to `'https://cdn.cocoapods.org/'`

PR #376, opened 2019-09-09
Swift Package Manager

ZevEisenberg:
Way of the future

PR #378, opened 2019-10-02
Add file to make BonMot installable through SPM

DomAmato:


ZevEisenberg:
@DomAmato thanks for your contribution! We actually already have a pull request open for Swift Package manager: #376. Would you mind looking at that and seeing if that would suit your needs? I’d like to get one of them merged soon, but I don’t use SPM, so your feedback would help me make sure we’re supporting it in a useful way.

DomAmato:
cool ill close this and give it a look. I made this change awhile ago and only got around to making a PR now haha

PR #381, opened 2019-10-31
Simplify StringStyle.update(part:) switch statement

### LINKED ISSUE #345: Use Swift 4.1's ability to wrap directives around switch cases
https://bugs.swift.org/browse/SR-2 was resolved in Swift 4.1.

Affects this code in StringStyle+Part.swift:

```swift
        case let .emphasis(emphasis):
            self.emphasis = emphasis
        default:
            // interaction between `#if` and `switch` is disappointing. This case
            // is in `default:` to remove a warning that default won't be accessed
            // on some platforms.
            switch stylePart {
            case let .hyphenationFactor(hyphenationFactor):
                self.hyphenationFactor = hyphenationFactor
            default:
                #if os(iOS) || os(tvOS) || os(watchOS)
                    switch stylePart {
                    case let .speaksPunctuation(speaksPunctuation):
                        self.speaksPunctuation = speaksPunctuation
                        return
```

erikstrottmann:
Fixes #345 using the improved interaction between `switch` and `#if` in Swift 4.1+.

I believe this requires removing support for Swift 4.0. Let me know if that assumption is incorrect.

### DIFF
--- a/Sources/StringStyle+Part.swift
+++ b/Sources/StringStyle+Part.swift
@@ -219,81 +219,52 @@ extension StringStyle {
             self.add(stringStyle: style)
         case let .emphasis(emphasis):
             self.emphasis = emphasis
-        default:
-            // interaction between `#if` and `switch` is disappointing. This case
-            // is in `default:` to remove a warning that default won't be accessed
-            // on some platforms.
-            switch stylePart {
-            case let .hyphenationFactor(hyphenationFactor):
-                self.hyphenationFactor = hyphenationFactor
-            default:
-                #if os(iOS) || os(tvOS) || os(watchOS)
-                    switch stylePart {
-                    case let .speaksPunctuation(speaksPunctuation):
-                        self.speaksPunctuation = speaksPunctuation
-                        return
-                    case let .speakingLanguage(speakingLanguage):
-                        self.speakingLanguage = speakingLanguage
-                        return
-                    case let .speakingPitch(speakingPitch):
-                        self.speakingPitch = speakingPitch
-                        return
-                    case let .speakingPronunciation(speakingPronunciation):
-                        self.speakingPronunciation = speakingPronunciation
-                        return
-                    case let .shouldQueueSpeechAnnouncement(shouldQueueSpeechAnnouncement):
-                        self.shouldQueueSpeechAnnouncement = shouldQueueSpeechAnnouncement
-                        return
-                    case let .headingLevel(headingLevel):
-                        self.headingLevel = headingLevel
-                        return
-                    default:
-                        break
-                    }
-                #endif
-
-                #if os(OSX) || os(iOS) || os(tvOS)
-                    switch stylePart {
-                    case let .numberCase(numberCase):
-                        self.numberCase = numberCase
-                    case let .numberSpacing(numberSpacing):
-                        self.numberSpacing = numberSpacing
-                    case let .fractions(fractions):
-                        self.fractions = fractions
-                    case let .superscript(superscript):
-                        self.superscript = superscript
-                    case let .`subscript`(`subscript`):
-                        self.`subscript` = `subscript`
-                    case let .ordinals(ordinals):
-                        self.ordinals = ordinals
-                    case let .scientificInferiors(scientificInferiors):
-                        self.scientificInferiors = scientificInferiors
-                    case let .smallCaps(smallCaps):
-                        self.smallCaps.insert(smallCaps)
-                    case let .stylisticAlternates(stylisticAlternates):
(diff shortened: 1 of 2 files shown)


ZevEisenberg:
I belieeeeve that’s correct, but I’d feel better if I could see a Swift bug # saying when it was changed. As for this update, it looks great, but I’m not sure I’m comfortable dropping Swift 4.0 support just yet. Xcode 11 won’t be required for App Store submissions [until April of 2020](https://developer.apple.com/app-store/submissions/), and even then, people may be using this for apps that are not in the App Store. The code is gross, yes, and your cleanup is really nice, but since the code itself is working and (mostly) well-tested, I see this as an ain’t-broke-don’t-fix-it scenario.

If any of our assumptions about Xcode versions are wrong, or just given enough time, I think we could reconsider.

ZevEisenberg:
Oh, and thanks for this contribution! The code itself looks really nice. It’s just complicated by some unfortunate historical/technical context.

erikstrottmann:
I should have linked directly to [SR-2](https://bugs.swift.org/browse/SR-2)—it was closed with a comment that it was implemented in Swift 4.1.

This was interesting to me, since apple/swift#9457 was merged into `master` on 2017-06-28, while [`swift-4.0-RELEASE`](https://github.com/apple/swift/releases/tag/swift-4.0-RELEASE) was tagged later than that on 2017-09-19 and [`swift-4.1-RELEASE`](https://github.com/apple/swift/releases/tag/swift-4.1-RELEASE) wasn’t tagged until 2018-03-29. So I confirmed via Git history that the `swift-4.0-RELEASE` commit does not have the SR-2 commit as an ancestor, but the `swift-4.1-RELEASE` commit does.

Instead of increasing the Swift version requirement from 4.0 to 4.2, it would be possible to also support 4.1, which is [included with Xcode 9.3](https://swift.org/blog/swift-4-1-released/). I can’t find the current minimum Xcode version for App Store submissions, but presumably it’s already Xcode 10. Still, you’re right that apps that aren’t on the App Store might be using BonMot with Xcode < 9.3.

I also noticed that [the readme badge mentions support for Swift 4.2 + 5.0](https://github.com/Rightpoint/BonMot/blame/51b98a8ba5526687c8a9f103c19127bca40ba5b1/README.md#L3). Should that include 4.0 too?

ZevEisenberg:
Good call on the badge. I’m not sure if it’s helping or hurting more. 4.2 and 5.0 are the versions we’re testing with on CircleCI (see .circleci/config.yml if you want the details), so in that sense, they’re all that’s currently supported. But that doesn’t mean it doesn’t work in older versions; just that we’re not currently guaranteeing compatibility. No one has complained yet, so maybe we’re good? ¯\\\_(ツ)\_/¯

erikstrottmann:
Hey there! Happy one year since opening this PR. (And what a year.) I wanted to check in and see what your thoughts are on getting this ready to merge. No pressure either way, but I figure at this point we should either merge or close this PR.

ZevEisenberg:
Hi! Yes, let's get this merged. Some stuff has changed on the build side of these things, so maybe you could rebase on latest upstream? That would kick off new builds. And I agree that the readme should be explicit about Swift 4.1+ or 4.2+ or whatever you deem appropriate based on the changes.

Sources/StringStyle+Part.swift
```diff
@@ -219,81 +219,52 @@ extension StringStyle {
...
-                        #endif
-                    }
-                #else
-                    fatalError("StylePart \(stylePart) should have been caught by an earlier case.")
-                #endif
-            }
+        case let .hyphenationFactor(hyphenationFactor):
+            self.hyphenationFactor = hyphenationFactor
+        #if os(iOS) || os(tvOS) || os(watchOS)
+        case let .speaksPunctuation(speaksPunctuation):
```
ZevEisenberg, line 225: I think I'd prefer to keep everything indented inside the `#if` blocks to make it easier to visually chunk this into its constituent parts.

erikstrottmann:
Oops, accidentally closed the PR. Reopened now.

I rebased on `master` and made the changes you requested.

I think it’s best to remove Swift 4.0 support and increase the minimum version to Swift 4.2, rather than adding new support for Swift 4.1.

Sources/StringStyle+Part.swift
erikstrottmann: Okay, I changed it. I originally removed that indentation because the default Xcode indentation (highlight code, then `control+I`) doesn’t add extra indentation within `#if` blocks.

erikstrottmann:
GitHub won’t let me reopen the PR because I force-pushed while it was closed. Huh.

Let’s continue in #396.

PR #382, opened 2019-11-15
Enable tests for SwiftPM

helje5:
A way to enable tests even when building w/ SwiftPM.

Only two tests fail (on macOS):
- testAvailableFeatures
- testImageTinting

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -14,18 +14,15 @@ let package = Package(
     products: [
         .library(
             name: "BonMot",
-            targets: ["BonMot"]),
+            targets: ["BonMot"])
     ],
     targets: [
         .target(
             name: "BonMot",
-            dependencies: [],
             path: "Sources"),
-        // Test targets do not currently support including resource files,
-        // which these tests depend on. Revisit when resources are supported.
-//        .testTarget(
-//            name: "BonMotTests",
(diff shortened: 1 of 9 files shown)


chrisballinger:
Thanks for the PR! This is an interesting and awesome workaround for the "SPM packages can't include resources" issue. [SE-0271](https://github.com/apple/swift-evolution/blob/master/proposals/0271-package-manager-resources.md) will resolve this problem at the source, so it might be better to put this on pause until resource bundles are officially supported in SPM.

chrisballinger:
Looks like this is resolved in Swift 5.3, would love to get this updated and merged once that's out: https://github.com/apple/swift-evolution/blob/master/proposals/0271-package-manager-resources.md

chrisballinger:
This should now be possible to do without hacks in Xcode 12.2 using the new Swift 5.3 SPM toolchain for bundled resources.

ZevEisenberg:
It looks like there's a bug related to bundled resources and test targets specifically. Not sure if this is going to be fixed in Xcode 12.5, but it doesn't seem to be in the betas. https://github.com/apple/swift-package-manager/pull/2905

chrisballinger:
Finally shipped in #411 #412

PR #389, opened 2020-04-14
Xcode 11.4 fixes

ZevEisenberg:


.circleci/config.yml
```diff
@@ -100,7 +100,7 @@ jobs:
 executors:
   xcode-11:
     macos:
-      xcode: "11.3.1"
+      xcode: "11.4"
```
ZevEisenberg, line 103: I replaced Xcode 11.3.1, but you might want to add 11.4 as a new Xcode, depending on how quickly people are adopting 11.4. There were some compiler crashes that made it a little scary.

chrisballinger, line 103: Not sure what happens if it's missing the ".0": https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions

```suggestion
      xcode: "11.4.0"
```

.circleci/config.yml
```diff
@@ -58,7 +58,7 @@ jobs:
       - setup
       - run: swift build
 
-  test-swift-5-0:
+  test-swift-5-2:
```
chrisballinger, line 61: Since this breaks the required CI status checks, maybe it'd be easier to keep as-is, since "5.0" really means "5.2", at least as far as `SWIFT_VERSION` is concerned

```suggestion
  test-swift-5-0:
```

.circleci/config.yml
```diff
@@ -149,7 +149,7 @@ workflows:
           filters:
             tags:
               only: /.*/
-      - test-swift-5-0:
+      - test-swift-5-2:
```
chrisballinger, line 152:
```suggestion
      - test-swift-5-0:
```

.circleci/config.yml
```diff
@@ -165,7 +165,7 @@ workflows:
           requires:
             - danger
             - build-swift-package
-            - test-swift-5-0
+            - test-swift-5-2
```
chrisballinger, line 168:
```suggestion
            - test-swift-5-0
```

Sources/StringStyle+Part.swift
```diff
@@ -172,10 +172,10 @@ extension StringStyle {
...
         case let .color(color):
             self.color = color
-        case let .underline(underline):
-            self.underline = underline
-        case let .strikethrough(strikethrough):
-            self.strikethrough = strikethrough
+        case let .underline(style, color):
+            self.underline = (style, color)
+        case let .strikethrough(style, color):
+            self.strikethrough = (style, color)
```
chrisballinger, line 178: What was the warning here? Does this have something to do with a change in implicit tuple behavior?

chrisballinger:
For the sake of getting this merged more quickly, I'd say maybe revert the gem update, since it looks like fixing that will take a while. I'll take a deeper dive for that in a separate PR.

ZevEisenberg:
I tried more conservatively updating just CocoaPods; we'll see if it's any happier now.

PR #391, opened 2020-05-04
New XMLStyleRule variants for dynamic styling

kosyak:
Hello and thank you for this fine library!

I propose addition of several `XMLStyleRule`s to dynamically generate and apply styles, enter- and exit-insertions of elements.

With these changes, you can use them like this:

```swift
// We want to style an xml that contains tags with some important attributes

var quoteAuthorStack = [String]()

let attributedString = try NSAttributedString.composed(
    ofXML: rawXml,
    rules: [
        .style("quote", quoteStyle),

        // We want to insert an image with a beautiful quote before the actual quote text.
        // Also we have an author's name in "author" attribute and we want 
        // to append this name after the quote text.
        .enterBlock(element: "quote") { attributes in
            // Saving author's name for later use
            quoteAuthorStack.append(attributes["author"]!)

            // "author" handling was hacky, so to show `attributes` in action,
            // why don't we have different quote images for different occasions...
            return NSAttributedString.composed(of: [ 
                UIImage(named: attributes["quotestyle"]!)!, 
                Special.lineFeed 
            ])
        },
        
        .exitBlock(element: "quote") {
            // Using our saved "author" attribute here to append the name after the quote
            let author = quoteAuthorStack.popLast()!

            return NSAttributedString.composed(of: [ 
                Special.lineFeed, author.styled(with: authorStyle) 
            ])
        },

        // Also we have a "link" tag with some attributes that will help us
        // to generate an URL
        .styleBlock("link") { attributes in
            guard let type = attributes["type"], [ "event", "person" ].contains(type),
                let id = attributes["id"],
                let link = URL(string: "\(Const.Domain)/\(type)/\(id)") else { return defaultStyle }

            return defaultStyle.byAdding(.link(link))
        }
    ]
)
```

We have a BonMot patched with these changes for quite some time, they are very convenient for us :)

### DIFF
--- a/Tests/AttributedStringStyleTests.swift
+++ b/Tests/AttributedStringStyleTests.swift
@@ -563,22 +563,14 @@ class StringStyleTests: XCTestCase {
 
         let rules: [XMLStyleRule] = [
             .styleBlock("one") { attributes in
-                guard let val1 = attributes["attr1"], let intVal1 = Int(val1) else {
-                    XCTFail("attr1 not found")
-                    return StringStyle()
-                }
+                tagAttr1Value = attributes["attr1"]
 
-                tagAttr1Value = val1
-                return StringStyle(.baselineOffset(CGFloat(intVal1)))
+                return StringStyle(.baselineOffset(CGFloat(Int(tagAttr1Value ?? "") ?? 0)))
             },
             .styleBlock("two") { attributes in
-                guard let val2 = attributes["attr2"], let intVal2 = Int(val2) else {
-                    XCTFail("attr2 not found")
-                    return StringStyle()
-                }
+                tagAttr2Value = attributes["attr2"]
 
-                tagAttr2Value = val2
-                return StringStyle(.baselineOffset(CGFloat(intVal2)))
+                return StringStyle(.baselineOffset(CGFloat(Int(tagAttr2Value ?? "") ?? 0)))
             },
         ]
 
@@ -614,39 +606,19 @@ class StringStyleTests: XCTestCase {
 
         let rules: [XMLStyleRule] = [
             .enterBlock(element: "one") { attributes in
-                guard let val1 = attributes["attr1"], let intVal1 = Int(val1) else {
-                    XCTFail("attr1 not found")
-                    return ""
-                }
+                tagAttr1Value = attributes["attr1"]
+                tagAttr2Value = attributes["attr2"]
 
-                guard let val2 = attributes["attr2"] else {
-                    XCTFail("attr2 not found")
-                    return ""
-                }
-
-                tagAttr1Value = val1
-                tagAttr2Value = val2
-
-                return val2.styled(with: .baselineOffset(CGFloat(intVal1)))
+                return (tagAttr2Value ?? "").styled(with: .baselineOffset(CGFloat(Int(tagAttr1Value ?? "") ?? 0)))
             },
             .exitBlock(element: "one") {
                 return "c"
             },
             .enterBlock(element: "two") { attributes in
-                guard let val3 = attributes["attr3"], let intVal3 = Int(val3) else {
-                    XCTFail("attr3 not found")
-                    return ""
-                }
+                tagAttr3Value = attributes["attr3"]
+                tagAttr4Value = attributes["attr4"]
 
-                guard let val4 = attributes["attr4"] else {
-                    XCTFail("attr4 not found")
-                    return ""
-                }
-
-                tagAttr3Value = val3
-                tagAttr4Value = val4
-
-                return val4.styled(with: .baselineOffset(CGFloat(intVal3)))
+                return (tagAttr4Value ?? "").styled(with: .baselineOffset(CGFloat(Int(tagAttr3Value ?? "") ?? 0)))
             },
             .exitBlock(element: "two") {
                 return "d"
--- a/Tests/AttributedStringStyleTests.swift
+++ b/Tests/AttributedStringStyleTests.swift
@@ -555,11 +555,11 @@ class StringStyleTests: XCTestCase {
         XCTAssertFalse(stillHasAltSixDict)
     }
 
-    func testStyleBlockRules() {
-        let string = "0<one attr1=\"11\">1<two attr2=\"12\">2</two></one>"
+    func testStyleBlockRules() throws {
+        let string = #"0<one attr1="11">1<two attr2="12">2</two></one>"#
 
-        var tagAttr1Value: String!
-        var tagAttr2Value: String!
+        var tagAttr1Value: String?
+        var tagAttr2Value: String?
 
         let rules: [XMLStyleRule] = [
             .styleBlock("one") { attributes in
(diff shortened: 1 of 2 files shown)


ZevEisenberg:
@kosyak wow, this is a great addition! Thanks so much for contributing back your work so that others can benefit ❤️

I'd love to see some unit tests to make sure it keeps working as advertised. Is that something you'd be willing to add? I can help if you run into trouble.

chrisballinger:
@kosyak This is great, thank you! I would also love to see the docs updated, with maybe an advanced/dynamic XML section explaining usage of this new functionality (similar to your PR description): https://github.com/Rightpoint/BonMot/blob/master/README.md#styling-parts-of-strings-with-xml

kosyak:
Thanks for your support! I totally understand the need of tests and docs for this, will come up with those, won’t hesitate to ask for help :)

kosyak:
@ZevEisenberg I've added unit tests (they're based on `testAdditiveFontFeatures` where I've found some code with `.xmlRules`). They may be in the wrong place or overcomplicated, please point me to the right direction.

Tests/AttributedStringStyleTests.swift
```diff
@@ -555,6 +555,134 @@ class StringStyleTests: XCTestCase {
         XCTAssertFalse(stillHasAltSixDict)
     }
 
+    func testStyleBlockRules() {
+        let string = "0<one attr1=\"11\">1<two attr2=\"12\">2</two></one>"
```
ZevEisenberg, line 559: [Raw strings](https://www.hackingwithswift.com/articles/162/how-to-use-raw-strings-in-swift) make this a lot nicer:

```suggestion
        let string = #"0<one attr1="11">1<two attr2="12">2</two></one>"#
```

Tests/AttributedStringStyleTests.swift
```diff
@@ -555,6 +555,134 @@ class StringStyleTests: XCTestCase {
...
+
+        var tagAttr1Value: String!
```
ZevEisenberg, line 561: No need for these to be `!`, since `XCTAssertEqual` does optional-promotion for you. i.e. this can be `String?`, and it will Do The Right Thing™

Tests/AttributedStringStyleTests.swift
ZevEisenberg: Also, I find the use of "11" and "12" in the mix to be confusing. I'd love longer and/or less ambiguous names, just to make it easier to read this code 🙂

Tests/AttributedStringStyleTests.swift
```diff
@@ -555,6 +555,134 @@ class StringStyleTests: XCTestCase {
...
+        var tagAttr2Value: String!
+
+        let rules: [XMLStyleRule] = [
+            .styleBlock("one") { attributes in
+                guard let val1 = attributes["attr1"], let intVal1 = Int(val1) else {
+                    XCTFail("attr1 not found")
+                    return StringStyle()
+                }
+
+                tagAttr1Value = val1
```
ZevEisenberg, line 571: I think you may be able to simplify this test. Rather than asserting things inside the closure, do some work in the closure, and afterward, asset that the result is as expected. Does that work for you, or would it miss some aspect of the way the closers work that you want to test?

Tests/AttributedStringStyleTests.swift
```diff
@@ -555,6 +555,134 @@ class StringStyleTests: XCTestCase {
...
+        XCTAssertEqual(attributed.string, "012")
+
+        XCTAssertEqual(tagAttr1Value, "11")
+        XCTAssertEqual(tagAttr2Value, "12")
+
+        let attrs0 = attributed.attributes(at: 0, effectiveRange: nil)
+        let attrs1 = attributed.attributes(at: 1, effectiveRange: nil)
+        let attrs2 = attributed.attributes(at: 2, effectiveRange: nil)
+
+        XCTAssertTrue(attrs0.count == 1)
```
ZevEisenberg, line 595:
```suggestion
        XCTAssertEqual(attrs0.count, 1)
```
For this and the two lines below

Tests/AttributedStringStyleTests.swift
```diff
@@ -555,6 +555,134 @@ class StringStyleTests: XCTestCase {
...
+        XCTAssertTrue(attrs1.count == 2)
+        XCTAssertTrue(attrs2.count == 2)
+
+        XCTAssertNil(attrs0[.baselineOffset])
+        guard let lineSpacing1 = attrs1[.baselineOffset] as? Int else { XCTFail("lineSpacing1 not found"); return }
```
ZevEisenberg, line 600: You should be able to mark the test method as `throws` and use `let lineSpacing1 = try XCTUnwrap(attrs1[.baselineOffset] as? Int)` to achieve the same effect, for this line and below.

Tests/AttributedStringStyleTests.swift
```diff
@@ -555,6 +555,134 @@ class StringStyleTests: XCTestCase {
...
+        let string = "0<one attr1=\"11\" attr2=\"a\">1<two attr3=\"12\" attr4=\"b\">2</two></one>"
+
+        var tagAttr1Value: String!
+        var tagAttr2Value: String!
+        var tagAttr3Value: String!
+        var tagAttr4Value: String!
+
+        let rules: [XMLStyleRule] = [
+            .enterBlock(element: "one") { attributes in
+                guard let val1 = attributes["attr1"], let intVal1 = Int(val1) else {
```
ZevEisenberg, line 617: Now that I understand a little more about what's going on, I think you can use `XCTUnwrap` to simplify all lines that have this pattern. Basically anywhere you currently have a `guard`, I think an unwrap would work better (and simpler) for tests.

One caveat is that you would probably have to annotate the various blocks with `throws`, and I'm not sure how much that would complicate their use in the rest of the app, so this may be invalid advice _inside_ the blocks, but still good elsewhere in the tests.

Tests/AttributedStringStyleTests.swift
kosyak: In this test, I've tried not only to extract and assert attribute values but also to have them included in the styling work.

So I've chosen one NSAttributeString parameter (`baselineOffset`) and tried to produce a style with a `baselineOffset` parsed from attribute's value in each of `styleBlock`s. Because of that, values of these attributes are numbers.

It doesn't matter for me what NSAttributedString style to use in the test, maybe there is a better fit instead of `baselineOffset`. E.g. we can have font names as attribute values and try to construct and return `.font()` style within blocks...

kosyak:
I've pushed fixes that hopefully will resolve some of your concerns.

> I'm wondering if we can come up with better names for the new cases?
I'm watching at some Kotlin code right now, and there's `lazyMessage` here. We can have
`case .lazyStyle(String, ([String: String]) -> StringStyle)` (this is good, I think), `.lazyEnter`, `.lazyExit` (not so good...)

Tests/AttributedStringStyleTests.swift
kosyak: I pushed another commit where all `guard`s are removed. A bit better, I think. Still, lots of typecasts and `??`...

ZevEisenberg:
Yeah, the Kotlin naming doesn't seem to fit too well here. What about `styler(String, ([String: String]) -> StringStyle)` to pair with the `style(String, StringStyle)` we already have? You could potentially even `typealias Styler = ([String: String]) -> StringStyle`.

kosyak:
I'm fine with `.styler`. Should we rename `.enterBlock`, `.exitBlock`?

kosyak:
> typealias Styler = ([String: String]) -> StringStyle

I'm not sure where to put it - should it be near `XMLStyleRule`?. It could be within XMLStyleRule but there's already `XMLStyleRule.Styler` https://github.com/Rightpoint/BonMot/pull/391/files#diff-c9b2f909282ce5a1468d61cb0337c704R136.

Also, won't there be a confusion between our freshly renamed `XMLStyleRule.styler` and `XMLStyleRule.Styler`?

Here's one more naming option: `.dynamicStyle`, `.dynamicEnter`, `.dynamicExit` (or even `.deferred...`) - just yo be sure we thought of them all :)

ZevEisenberg:
Wow, I'm glad one of the two of us is paying attention! 😅 I haven't worked on this code in a while, so I'm a little rusty. You make excellent points. I'll have a think about it when I've got some time, and I definitely hope to get this in soon.

chrisballinger (approved):
@kosyak Thanks so much! I think where you landed on the naming makes sense to me. To avoid having this sit open any longer I'm going to merge it as-is, but it would be great if you could follow up with a docs update PR to the README!

kosyak:
Thanks for the merge, will work on readme/docs this week.

PR #395, opened 2020-11-03
Use Xcode 12 on CircleCI and CocoaPods environment context

chrisballinger:
Using CircleCI Environment Contexts: https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-context


colejd (approved):
Self-approving!

PR #398, opened 2020-12-17
Fixes UISegmentedControl extension crash and improves efficiency

### LINKED ISSUE #397: Crash in UISegmentedControl extension
There is a typo inside of the method `bon_titleTextAttributes(for:)`. When iterating over the existing `titleTextAttributes` it is assigning the attribute key/value pair as the new value rather than the value itself. This causes an unrecognized selector exception to be thrown and crashes the app.

To reproduce this crash you need to set some custom title text attributes on a segmented control and then either call `segmentedControl.adaptText(forTraitCollection:)` or turn on the adaptive content size monitor and change the content size in Settings > Accessibility > Larger Text.

rzulkoski:
Fixes #397

### DIFF
--- a/Sources/UIKit/AdaptableTextContainer.swift
+++ b/Sources/UIKit/AdaptableTextContainer.swift
@@ -147,7 +147,7 @@ extension UISegmentedControl {
     // `UISegmentedControl` has terrible generics ([NSObject: AnyObject]? or [AnyHashable: Any]?) on
     /// `titleTextAttributes`, so use a helper in Swift 3+
     @nonobjc final func bon_titleTextAttributes(for state: UIControl.State) -> StyleAttributes {
-        let attributes = titleTextAttributes(for: state)
+		let attributes = titleTextAttributes(for: state) ?? [:]
         #if swift(>=4.2)
             return attributes
         #else
--- a/Sources/UIKit/AdaptableTextContainer.swift
+++ b/Sources/UIKit/AdaptableTextContainer.swift
@@ -146,20 +146,20 @@ extension UISegmentedControl {
 
     // `UISegmentedControl` has terrible generics ([NSObject: AnyObject]? or [AnyHashable: Any]?) on
     /// `titleTextAttributes`, so use a helper in Swift 3+
-    @nonobjc final func bon_titleTextAttributes(for state: UIControl.State) -> StyleAttributes {
-        let attributes = titleTextAttributes(for: state) ?? [:]
-        var result: StyleAttributes = [:]
-        for value in attributes {
-            #if swift(>=4.2)
-                result[value.key] = value
-            #else
-                guard let string = value.key as? StyleAttributes.Key else {
-                    fatalError("Can not convert key \(value.key) to String")
-                }
-            result[string] = value
-            #endif
-        }
-        return result
+    @nonobjc final func bon_titleTextAttributes(for state: UIControl.State) -> StyleAttributes? {
+		guard let attributes = titleTextAttributes(for: state) else { return nil }
+		#if swift(>=4.2)
+			return attributes
+		#else
+			var result: StyleAttributes = [:]
+			for attribute in attributes {
+				guard let string = attribute.key as? StyleAttributes.Key else {
+					fatalError("Can not convert key \(attribute.key) to String")
+				}
+				result[string] = attribute.value
+			}
+			return result
+		#endif
     }
 
     /// Adapt `attributedTitle`, for all control states, to the specified trait collection.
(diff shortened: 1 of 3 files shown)


chrisballinger (commented):
Thanks! Would you mind reverting the tabs/spaces change?

rzulkoski:
@chrisballinger yes, of course! Sorry about that, it looked much better in Xcode, I promise. 😉

Sources/UIKit/AdaptableTextContainer.swift
```diff
@@ -146,20 +146,20 @@ extension UISegmentedControl {
...
-    @nonobjc final func bon_titleTextAttributes(for state: UIControl.State) -> StyleAttributes {
-        let attributes = titleTextAttributes(for: state) ?? [:]
-        var result: StyleAttributes = [:]
-        for value in attributes {
-            #if swift(>=4.2)
-                result[value.key] = value
-            #else
-                guard let string = value.key as? StyleAttributes.Key else {
-                    fatalError("Can not convert key \(value.key) to String")
+    @nonobjc final func bon_titleTextAttributes(for state: UIControl.State) -> StyleAttributes? {
```
chrisballinger, line 149: This changes the public API, would it be possible keep this the same?

chrisballinger (commented):
Thank you! Would you also mind taking a look at the test failures on CI?

Sources/UIKit/AdaptableTextContainer.swift
rzulkoski: Sure, I have reverted the changes to the signature of that method.

Sources/UIKit/AdaptableTextContainer.swift
```diff
@@ -147,19 +147,19 @@ extension UISegmentedControl {
     // `UISegmentedControl` has terrible generics ([NSObject: AnyObject]? or [AnyHashable: Any]?) on
```
ZevEisenberg, line 147: Does this comment need to be updated at all in light of these changes?

ZevEisenberg (approved):
Is there a unit test we could write to confirm the new behavior?

Sources/UIKit/AdaptableTextContainer.swift
rzulkoski: I fixed an error inside that method. When I changed the return type back to being non-optional I forgot to default to an empty dictionary when nil. I also added a test for segmented control similar to the other UIKit tests.

As for the comment, I can't speak to whether it is still appropriate or not, although I can say in Swift 5 `titleTextAttributes(for:)` returns `[NSAttributedFont.Key: Any]?`.

Sources/UIKit/AdaptableTextContainer.swift
rzulkoski: I just realized that method is actually `internal`, so if we wanted to keep that change it wouldn't affect the public API. I'm fine with either direction.

rzulkoski: Also it looks like the tests are failing because of a warning. It's either because Zev's twitter account is unreachable or there is a placeholder team ID in the 'App' target.

```
 ** BUILD SUCCEEDED **
    
   Testing with `xcodebuild`. 
 -> BonMot (5.6.0)
    - WARN  | url: The URL (https://twitter.com/ZevEisenberg) is not reachable.
    - NOTE  | xcodebuild:  note: Using new build system
    - NOTE  | xcodebuild:  note: Building targets in parallel
    - NOTE  | xcodebuild:  note: Using codesigning identity override: -
    - NOTE  | xcodebuild:  note: Planning build
    - NOTE  | xcodebuild:  note: Constructing build description
    - NOTE  | xcodebuild:  warning: Capabilities for Signing & Capabilities may not function correctly because its entitlements use a placeholder team ID. To resolve this, select a development team in the App editor. (in target 'App' from project 'App')
    - NOTE  | xcodebuild:  warning: Skipping code signing because the target does not have an Info.plist file and one is not being generated automatically. (in target 'App' from project 'App')
    - NOTE  | xcodebuild:  note: Using codesigning identity override: 

[!] BonMot did not pass validation, due to 1 warning (but you can use `--allow-warnings` to ignore it).
You can use the `--no-clean` option to inspect any issue.

Exited with code exit status 1
CircleCI received exit code 1
```

rzulkoski: Might be related to #373 @ZevEisenberg

rzulkoski:
@ZevEisenberg I commented out the social_media_url in the podspec and now all of the tests are passing. Any chance we could get this merged at some point? Thanks!

PR #400, opened 2020-12-18
Fixes notifyContainedAdaptiveContentSizeContainers()

### LINKED ISSUE #399: Views not in the view hierarchy using wrong trait collection
When flooding the view hierarchy with `notifyContainedAdaptiveContentSizeContainers()` in response to the `UIContentSizeCategory.didChangeNotification` notification there is an assumption that a view controllers view that is loaded but not yet in the view hierarchy will have the update trait collection available.

This is NOT the case.

The view controller does have the new trait collection, but its view and subviews will not have the new trait collection at this point.

rzulkoski:
Fixed UIView extension method notifyContainedAdaptiveContentSizeContainers() by adding the ability to specify the trait collection and supplying one when updating a view controllers views and subviews when they aren't in the view hierarchy yet.

Fixes #399


chrisballinger (approved):
Thanks! Looks like the CI failures were just related to that Twitter issue.

PR #405, opened 2021-03-14
Test cleanup and typo fixes

ZevEisenberg:
The only interesting thing (and calling it _interesting_ is probably a stretch) is the commit that fixes a typo in public API by deprecating and renaming it. Please let me know if you think I've done it in a good way, and whether it constitutes a breaking change. (If the latter, we might as well just rename it and skip the deprecation.) It's a tricky case because the renamed parameter has a defaulted value, so consumers may not even be aware that there was a typo.


chrisballinger (approved):
Thank you! I think the breaking change is minor enough to not need a major version bump, but we should at least bump the minor version.

PR #406, opened 2021-03-14
Fix enum for module stability

NikolayJuly:
For working project, I need to build this pod as binary framework. To avoid rebuild it with every new xcode, we build them with `BUILD_LIBRARY_FOR_DISTRIBUTION=YES`.

In this mode, swift generate swiftinterface file, which is crucial for module stability

Unfortunately it has bug, which make swiftinterface invalid for BonMot specifically.

Here is link for swift bug: https://bugs.swift.org/browse/SR-14355

Looks like it just skip string literals, because they are not needed,  and do not check that `RawValue` type is not `String`.

I see no other way, rather than change raw type to String. In this way swiftinterface can be compiled.

Additionally I added `CaseIterable` and still keep `all ` for backward compatibility.

I would appreciate if you take my changes and create new release, so I would avoid creating fork of BonMot repo

### DIFF
--- a/Sources/Special.swift
+++ b/Sources/Special.swift
@@ -87,6 +87,7 @@ extension Special {
 
     /// All of the enum values contained in `Special`.
     /// Property kept here for backward compatibility
+    @available(*, deprecated, renamed: "allCases")
     @inlinable
     public static var all: [Special] { allCases }
 
--- a/Sources/Special.swift
+++ b/Sources/Special.swift
@@ -10,7 +10,7 @@
 /// in `Special` are either non-printing (like the various space characters) or
 /// visually ambiguous when viewed with a monospace code font (like the dashes
 /// and hyphens).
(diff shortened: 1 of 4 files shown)


Sources/Special.swift
```diff
@@ -86,33 +86,8 @@ extension Special {
...
-        .paragraphSeparator,
-        .leftToRightOverride,
-        .narrowNoBreakSpace,
-        .wordJoiner,
-        .minusSign,
-        .objectReplacementCharacter,
-    ]
+    /// Property kept here for backward compatibility
+    @inlinable
+    public static var all: [Special] { allCases }
```
ZevEisenberg, line 91: Can we mark this as deprecated?

NikolayJuly: Good point, done

chrisballinger (approved):
I dig it. Thank you!

PR #411, opened 2021-06-08
Xcode 13 compatibility

### LINKED ISSUE #409: Compilation issue in Xcode 13 beta 1
AdaptiveStyle.swift:

```swift
/// Enable automatic scaling of fonts obtained using `UIFontMetrics`
/// available on iOS 11+ based on the provided `textStyle` and optional
/// `maxPointSize`. If `maxPointSize` is `nil` the font will grow unbounded.
@available(iOS 11, tvOS 11, *)
case fontMetrics(textStyle: BonMotTextStyle, maxPointSize: CGFloat?)
```

Error:

    🛑 Enum cases with associated values cannot be marked potentially unavailable with '@available'

ZevEisenberg:
Fixes #409, but does not add Xcode 13 to the CI test runners because CircleCI doesn't support Xcode 13 yet.


Sources/Platform.swift
```diff
@@ -27,8 +27,8 @@
     public typealias BONFontDescriptor = UIFontDescriptor
     public typealias BONSymbolicTraits = UIFontDescriptor.SymbolicTraits
     let BONFontDescriptorFeatureSettingsAttribute = UIFontDescriptor.AttributeName.featureSettings
-    let BONFontFeatureTypeIdentifierKey = UIFontDescriptor.FeatureKey.featureIdentifier
-    let BONFontFeatureSelectorIdentifierKey = UIFontDescriptor.FeatureKey.typeIdentifier
+    let BONFontFeatureTypeIdentifierKey = UIFontDescriptor.FeatureKey.type
+    let BONFontFeatureSelectorIdentifierKey = UIFontDescriptor.FeatureKey.selector
```
ZevEisenberg, line 31: These have been deprecated since iOS 8, but we somehow didn't get compile-time warnings, but in this release they must have finally removed them because we get run-time crashes due to failed dynamic symbol loading or something like that. Migrated them to the iOS 9+ variants.

Sources/UIKit/AdaptiveStyle.swift
```diff
@@ -10,37 +10,74 @@
...
 /// A few default font scaling behaviors.
-public enum AdaptiveStyle {
+public struct AdaptiveStyle {
+
+    public enum Behavior {
+        case control
+        case body
+        case preferred
+
+        @available(iOS 11, tvOS 11, *)
```
ZevEisenberg, line 20: I checked, and apparently the stuff we use for this actually goes back to iOS 10, but I don't see any point in back-porting it now.

Sources/UIKit/AdaptiveStyle.swift
```diff
@@ -10,37 +10,74 @@
...
+        self.textStyle = textStyle
+        self.maxPointSize = maxPointSize
+    }
 
     /// Scale the font up or down based on the Dynamic Type slider, but do not
     /// grow into the Accessibility ranges.
-    case control
+    public static var control: AdaptiveStyle {
+        AdaptiveStyle(behavior: .control)
+    }
```
ZevEisenberg, line 44: Converted all these enum cases to static functions to maintain call site compatibility.

ZevEisenberg:
Hmmm, it looks like there's some weirdness with `UIFontDescriptor.FeatureKey.type` and `UIFontDescriptor.FeatureKey.selector` in terms of iOS version checking. I can't figure out what incarnation I need. I thought this would work, but it does not:

```swift
let BONFontFeatureTypeIdentifierKey: UIFontDescriptor.FeatureKey = {
    if #available(iOS 10, macCatalyst 13, tvOS 10, watchOS 2, *) {
        return .type
    }
    else {
        return .featureIdentifier
    }
}()

let BONFontFeatureSelectorIdentifierKey: UIFontDescriptor.FeatureKey = {
    if #available(iOS 10, macCatalyst 13, tvOS 10, watchOS 2, *) {
        return .selector
    }
    else {
        return .typeIdentifier
    }
}()
```

That builds in Xcode 13, but in Xcode 12 I get:

    🛑 Type 'UIFontDescriptor.FeatureKey' has no member 'type'

chrisballinger:
Thanks @ZevEisenberg! Any further thoughts on why the availability check is not working as expected?

ZevEisenberg:
Nope. Not sure if bug or what. You might want to try reproing in a sample project, and then doing a WWDC lab or at least radar if you can't figure it out.

raven:
From my reading, it appears that the `@available` checks for enum cases with associated values was originally broken, and they have 'resolved' it in Swift 5.5

https://github.com/apple/swift/pull/36327

```
The old behavior:
 - Stored properties could not have @available at all.
 - Enum elements with associated values could be @available without restriction.

When computing the in-memory layout of a type we require that all stored types are available at runtime, that is, not any newer than the intersection of the current deployment target and the availability context of their type.

This means that stored properties and enum elements with associated values must have the same availability checking behavior as each other, and also that both existing behaviors were wrong.

The new rule is that @available is only allowed on either kind of declaration if the OS version is not any newer than the enclosing availability context.
```

chrisballinger:
The easiest and most minimal solution might be to bump the minimum iOS version to 11.0 and remove the availability check, which I think I might prefer at this point.

ZevEisenberg:
I'm fine with that. Maybe do an audit for any other availability checks that can be removed. It'll need to be a major version bump because of semver

xezero:
Any update on this? :)

mihaicris-adoreme:
Hi, is there any branch for BonMot that I can test my project on Xcode 13/iOS15 ?

Sources/Platform.swift
chrisballinger: There is a radar for this in the release notes: https://developer.apple.com/documentation/xcode-release-notes/xcode-13-beta-release-notes

> Apps that contain Swift code referencing featureIdentifier or typeIdentifier fail to launch on earlier OS releases. (79090498)

Seems like they might be fixing it at some point.

chrisballinger:
Sorry folks, looks like we are blocked on known bug in Xcode 13b2, unless y'all can come up with a workaround: https://github.com/Rightpoint/BonMot/pull/411#discussion_r659075502

chrisballinger (approved):

PR #426, opened 2023-01-15
composed removed []

DevVenusK:
[] can be removed when using the composed method.
Overloading methods for optional use.

### DIFF
--- a/Sources/Composable.swift
+++ b/Sources/Composable.swift
@@ -131,7 +131,7 @@ extension NSAttributedString {
     /// the `Composable` item only if the `Composable` does not have a style
     /// value configured.
     ///
-    /// - parameter composables: An array of `Composable` to join into an
+    /// - parameter composables: A list of `Composable` to join into an
     ///                          `NSAttributedString`.
     /// - parameter baseStyle: The base style to apply to every `Composable`.
     ///                        If no `baseStyle` is supplied, no additional
--- a/Sources/Composable.swift
+++ b/Sources/Composable.swift
@@ -125,20 +125,22 @@ extension NSAttributedString {
         return string
     }
     
+    /// Compose an `NSAttributedString` by concatenating every item in
+    /// `composables` with `baseStyle` applied. The `separator` is inserted
+    /// between every item. `baseStyle` acts as the default style, and apply to
+    /// the `Composable` item only if the `Composable` does not have a style
+    /// value configured.
+    ///
+    /// - parameter composables: An array of `Composable` to join into an
+    ///                          `NSAttributedString`.
+    /// - parameter baseStyle: The base style to apply to every `Composable`.
+    ///                        If no `baseStyle` is supplied, no additional
+    ///                        styling will be added.
+    /// - parameter separator: The separator to insert between every pair of
+    ///                        elements in `composables`.
+    /// - returns: A new `NSAttributedString`.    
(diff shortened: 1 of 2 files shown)


ZevEisenberg:
@DevVenusK thanks for your contribution! I don't fully understand the issue that this fixes. Would you mind posting a reproduction case? For bonus points, and unit test would be great as well.

DevVenusK:
Currently we have to use it like this to use the compose method.
`.composed(of: [ 
"apple".styled(...),
"is delicious".styled(...)
])`

However, if you use the added method, you can do this.
`.composed(of: 
"apple".styled(...),
"is delicious".styled(...)
)`

I really like the missing [].

ZevEisenberg:
Ah, now I understand. Can you perhaps simplify the code by calling through to an existing function, passing along the array from the variadic parameters, instead of duplicating the whole function body?

DevVenusK:
good idea! I fixed it!
Would you mind posting a description?

ZevEisenberg:
A description of what?

DevVenusK:
Sorry, I meant documentation comments, not a description. Is it okay to use the documentation comments of this method as is?

Sources/Composable.swift
```diff
@@ -124,6 +124,24 @@ extension NSAttributedString {
...
         return string
     }
+    
+    /// Compose an `NSAttributedString` by concatenating every item in
+    /// `composables` with `baseStyle` applied. The `separator` is inserted
+    /// between every item. `baseStyle` acts as the default style, and apply to
+    /// the `Composable` item only if the `Composable` does not have a style
+    /// value configured.
+    ///
+    /// - parameter composables: An array of `Composable` to join into an
```
ZevEisenberg, line 134: Not an array any more

ZevEisenberg (changes requested):
If the comment is copy-and-pasted, it should be good, with the exception of any mention of an array, which should probably be changed to "list"

DevVenusK:
Oh, I hadn't thought of that thanks. Edited.

ZevEisenberg:
Looking good! I'd still like to see a unit test to make sure we don't break this syntax in future.

DevVenusK:
Write unit test for variadic parameter. 🙇🏻‍♂️

Tests/ComposableTests.swift
```diff
@@ -18,7 +18,7 @@ import AppKit
 #if canImport(UIKit) || canImport(AppKit)
 
 class ComposableTests: XCTestCase {
-
+
```
ZevEisenberg, line 21: Please remove extra whitespace

ZevEisenberg (changes requested):

ZevEisenberg:
Nice test! Let's get the whitespace cleaned up and then we're good to merge.

DevVenusK:
> Nice test! Let's get the whitespace cleaned up and then we're good to merge.

done!

ZevEisenberg (approved):

DevVenusK:
@ZevEisenberg Do you have any idea when this PR will be merged?

ZevEisenberg:
Sorry, I've been procrastinating on it because I don't know if our CocoaPods release process works any more. I'll see what I can do.

ZevEisenberg:
Merged, but having trouble with the CocoaPods release, as anticipated. Working on it.

DevVenusK:
thx!! 🤩

vpavelRP:
@DevVenusK @ZevEisenberg THe cocoapods 6.1.2 is published as well

PR #433, opened 2023-08-15
Fix xcode 15 warnings

### LINKED ISSUE #432: Xcode 15 Warnings
Hi Guys!
I compiled your library in `Xcode 15.0 Beta 6 (15A5219j)` and received the following warnings:

Screenshot 2023-08-15 at 15 59 16

`Forming 'UnsafeMutableRawPointer' to a variable of type 'NSString'; this is likely incorrect because 'NSString' may contain an object reference.`

Could you please fix this in future releases?
You can find information about this new warning and suggested solutions in [proposal](https://github.com/atrick/swift-evolution/blob/diagnose-implicit-raw-bitwise/proposals/nnnn-implicit-raw-bitwise-conversion.md#associated-object-string-keys)

ZevEisenberg:
Fixes #432


ZevEisenberg:
There's a bunch of churn in these commits, because I'm removing tooling that no one has time or inclination to maintain. The real reason for the failure seems to be some issues with line/column error reporting with XMLParser in some recent version of Apple's OSes. At some point I'll either track it down or disable those tests.

PR #434, opened 2023-08-15
Apple Silicon CI Resource Class

eliotw1:
Based on https://circleci.com/blog/m1-mac-resource-class/


.circleci/ruby_setup.sh
```diff
@@ -0,0 +1,47 @@
...
+    echo "Installing ruby-build..."
+    brew install ruby-build
+else
+    echo "ruby-build is already installed."
+fi
+
+# Ensure ruby-build is up-to-date
+brew update && brew upgrade ruby-build
+
+# Check if Ruby 3.0.6 is already installed; if not, install it
```
ZevEisenberg, line 37: We reference Ruby 3.0.6 a bunch of times in this file and the next. Can we use a standard .ruby-version file in the project, which rbenv will respect, so we don't need to manually reference it everywhere?

ZevEisenberg:
One of the failures we're seeing is surrounding a test we have for the error messages we get from XMLParser. When it encounters an error, we try to be helpful and supply the column and line in the XML string where the error occurred. We do that by looking at the column and line where the parser is when we log the error. In Xcode 14 and aligned releases, the column number just seems to be wildly wrong. Here's someone who saw something similar: https://stackoverflow.com/questions/7137322/nsxmlparser-column-number-is-wrong

.circleci/config.yml
```diff
@@ -109,14 +115,47 @@ executors:
     shell: /bin/bash --login -eo pipefail
   xcode-13:
```
ZevEisenberg, line 116: I think we can safely ditch Xcode 13 and just test 14.

.circleci/ruby_setup.sh
```diff
@@ -0,0 +1,47 @@
...
+if ! rbenv versions | grep -q 3.0.6; then
```
ZevEisenberg, line 38: We might be able to avoid a lot of (all of?) this script because it looks like Circle has support for switching between modern rubies: https://circleci.com/docs/testing-ios/#using-ruby

PR #437, opened 2024-03-26
Added Privacy Manifest for iOS

KaustubhKulkarni2202:
- Adds a privacy manifest.
- Updates Package.swift to include the privacy manifest
- Updates podspec to include the privacy manifest

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -21,7 +21,7 @@ let package = Package(
             name: "BonMot",
             dependencies: [],
             path: "Sources",
-            exclude: ["Info.plist"]
+            exclude: ["Info.plist"],
             resources: [
                 .copy("PrivacyInfo.xcprivacy"),
         ]),
--- a/Package.swift
+++ b/Package.swift
@@ -22,7 +22,9 @@ let package = Package(
             dependencies: [],
             path: "Sources",
             exclude: ["Info.plist"]
-        ),
+            resources: [
+                .copy("PrivacyInfo.xcprivacy"),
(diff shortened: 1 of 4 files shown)


ZevEisenberg:
Thanks! Maybe this will be the push I need to get our CI working again. I have an old WIP PR at https://github.com/Rightpoint/BonMot/pull/433 that I might dust off.

Package.swift
```diff
@@ -35,6 +37,7 @@ let package = Package(
             ],
             resources: [
                 .process("Resources"),
+                .copy("PrivacyInfo.xcprivacy"),
```
ZevEisenberg, line 40: Is this needed in the test target?

Package.swift
```diff
@@ -21,8 +21,10 @@ let package = Package(
             name: "BonMot",
             dependencies: [],
             path: "Sources",
-            exclude: ["Info.plist"]
-        ),
+            exclude: ["Info.plist"],
+            resources: [
+                .copy("PrivacyInfo.xcprivacy"),
```
ZevEisenberg, line 26: Why `copy` instead of `process`?

KaustubhKulkarni2202: As we are not applying a platform-specific rule to the resource we need copy of it.

Package.swift
kapitoshka438: No

KaustubhKulkarni2202: not needed

PR #438, opened 2024-07-10
Apply fixes to warnings with complete concurrency warnings

AgapovOne:
Applied not all fixes, but it is a start.

We have warning on emphasis now so it would be great to have it sendable.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,6 +127,9 @@ commands:
     description: "Restore Ruby Gems"
     steps:
       - run:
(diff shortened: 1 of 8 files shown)


AgapovOne:
CI is a mess

ZevEisenberg:
You're tellin' me! I have a branch in progress to massively simplify CI, but I ran into boring test issues that I haven't had the energy to fix. At this point you're likely going to want to fork because I don't know if there's anyone with the time and expertise to maintain this library in public any more 😔

PR #439, opened 2024-07-19
Workspace for CI

AgapovOne:


AgapovOne:
Seems like circle ci just blocked me :)

ZevEisenberg:
Haha oops. I forget what I was doing on my other branch (it looks like you're using some of it though?), but I was thinking of moving over to GitHub actions. Also dropping old Xcode versions; too much of a headache, and folks are usually better about upgrading these days.

AgapovOne:
Yes, I just used your work and wanted to try more. 

I see that issue is Resources with swift test command. It just breaks with bundle retrieval of images. 

I will try to convert to GitHub actions

ZevEisenberg:
Thanks! I definitely regret putting so much image stuff in this project; what a headache! I'm also tinkering with what I hope could become BonMot's successor: a new library that allows customized styling of `AttributedString` via Markdown using the same underlying markdown stack as Foundation is using.

AgapovOne:
I did a github actions integration. Here is a build: https://github.com/AgapovOne/BonMot/actions/runs/10029740863?pr=1

- Added xcresult of tests
- Added upload artifacts step for easier investigation

Now I don't get why so many tests are broken. Tried to fix bundle issues with image retrieval on macOS, failed.

Then tried to see what happened with dynamic type on tvOS. Seems like Apple just removed any dynamic type support and they have only Large now?

